Issue #793, opened 2015-04-24
Mixed Content: web player breaks on https / SSL (loads insecure resources) 


patrickhunt:
Mixed Content errors break the web player when viewing over https.

**To Reproduce**:
Visit any game and put https in the url, (example: https://play.textadventures.co.uk/Play.aspx?id=fvjvccvef0mv0dyl-urzsg ).  Currently this causes the page get stuck on "Loading..."  The Mixed Content errors can be seen in the developer tools console.

**To Fix**:
All web requests to external resources should use https, not http.  See http://www.paulirish.com/2010/the-protocol-relative-url/ for more.

PR #794, opened 2015-04-24
First shot at moving any http external resources to https.

patrickhunt:
Fixes Issue #793

I can't test, so there might be some that I missed.

### DIFF
--- a/WebEditor/WebEditor/Views/Shared/_Layout.cshtml
+++ b/WebEditor/WebEditor/Views/Shared/_Layout.cshtml
@@ -3,14 +3,14 @@
 <head>
     <meta charset="utf-8" />
     <title>@ViewBag.Title</title>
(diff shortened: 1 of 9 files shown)


alexwarren:
Would it not be better to use scheme-relative URLs instead of hardcoding as https?

e.g.

    <link rel="Stylesheet" type="text/css" href="//ajax.aspnetcdn.com/ajax/jquery.ui/1.10.2/themes/redmond/jquery-ui.css" />

patrickhunt:
Apparently the new recommendation is to use https whenever possible, (http://www.paulirish.com/2010/the-protocol-relative-url/)  But yeah, the protocol agnostic '//example.com' would work fine.

alexwarren:
Thanks, I finally merged this (manually, as there were conflicts because I'd left this open so long).

PR #828, opened 2015-12-15
added notfirst functionality to text processor in coreoutput.aslx

F2Andy:
This commit adds the ability to only output text if this is not the first time to the text processor, following a request on the forum.

Added a new function "ProcessTextCommand_NotFirst", which will return an empty string the first time, and the given string any other time. The function is basically "ProcessTextCommand_Once" with the last couple of lines swapped around (kind of).

Added three lines to "ProcessTextCommand", which will conditionally invoke "ProcessTextCommand_NotFirst".

No tests included, but I have tested it for my own game.


alexwarren:
Thanks!

PR #889, opened 2017-02-15
template update

Pertex:
Update of deutsch.aslx
I added some links to the allfunctions page of the wiki. Could you pull this page so that I can see if it works?


ThePix:
Hi Pertex


I have slightly changed the language support for wearables. The big thing is that it add (worn) to an item's alias, and I forgot to use a template for that, so there is a new template:


<template name="wornmodifier">worn</template>

Also, three other used object.display, but are better using GetDisplayGarment(object) as that will give the correct prefix and suffix. If you send me the German for "worn", I can update the file, if that is easier.

Andy


________________________________

Pertex:
Hi Andy,
I would say "angezogen". A bit long but I cant find another word for it

Pertex

PR #890, opened 2017-03-04
update docu all functions

Pertex:
added missing functins to the all functions
page


ThePix:
Combines functions and script commands into a single index.

PR #923, opened 2017-11-16
update

Pertex:
-reading data from external server/file
- fixing languagefiles


KVonGit:
Hello, Pertex!

What does this allow us to do?

...or how do we use it?  (That's the real question.)

PR #931, opened 2018-02-02
Adding the function ASLAttribute()

KVonGit:
ASLAttribute() is identical to ASLEvent(), but it does not call TryFinishTurn().

It can be used to pass values from JS to Quest without causing the turn to finish prematurely.


KVonGit:
Rewriting this code and sending a new pull request.

PR #933, opened 2018-02-06
Adding the function: ASLCall()

KVonGit:
Adding ```ASLCall()``` and ```SendCall()```.

These work together.

```ASLCall()``` is just like ```ASLEvent()```, except it does not call ```TryFinishTurn()```.

---
I added a bit of code for this under each bit of code concerning ```ASLEvent()```.  Even in **LegacyGame.vb**.  

I was leery of adding it there, but I decided to let it rock if it passed all the tests.

...and it passed all the tests:

image

I don't know if that could effect the playability of legacy games.  I'm sure passing those tests doesn't mean it's sure to work.  

...but I tested it out by playing The Mansion in a debug build, and the game seemed to work fine.

### DIFF
--- a/IASL/IASL.cs
+++ b/IASL/IASL.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
@@ -17,6 +17,7 @@ namespace TextAdventures.Quest
         void SendCommand(string command);
         void SendCommand(string command, IDictionary<string, string> metadata);
         void SendEvent(string eventName, string param);
+        void SendCall(string eventName, string param);
         event PrintTextHandler PrintText;
         event UpdateListHandler UpdateList;
         event FinishedHandler Finished;
--- a/Legacy/LegacyGame.vb
+++ b/Legacy/LegacyGame.vb
@@ -7839,9 +7839,9 @@ Public Class LegacyGame
         ' beginning or end of a line.
         ' PostScript: well, it used to, I'm not sure if it's really
         ' required now though....
-        ' As of Quest 4.0 we use the � character rather than a dot.
-        test = "�" & Trim(test) & "�"
-        required = "�" & required & "�"
+        ' As of Quest 4.0 we use the ¦ character rather than a dot.
+        test = "¦" & Trim(test) & "¦"
+        required = "¦" & required & "¦"
 
         'Go through RequiredLine in chunks going up to variables.
         Dim currentReqLinePos = 1
--- a/Legacy/LegacyGame.vb
+++ b/Legacy/LegacyGame.vb
@@ -7839,9 +7839,9 @@ Public Class LegacyGame
         ' beginning or end of a line.
         ' PostScript: well, it used to, I'm not sure if it's really
         ' required now though....
-        ' As of Quest 4.0 we use the ¦ character rather than a dot.
-        test = "¦" & Trim(test) & "¦"
-        required = "¦" & required & "¦"
+        ' As of Quest 4.0 we use the ï¿½ character rather than a dot.
+        test = "ï¿½" & Trim(test) & "ï¿½"
+        required = "ï¿½" & required & "ï¿½"
 
         'Go through RequiredLine in chunks going up to variables.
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by KVonGit** [comment]
I believe I've found a better way to do this.  

Only one core function needs to be modified, and there is one new function to work with it.

http://textadventures.co.uk/forum/games/topic/nvwbfy2c8k6gj-xjjghxda/setokaivas-collection-of-novelized-roleplaying-logs-v1-5

<details>

Here is the modified FinishTurn() function:

```xml
  <function name="FinishTurn">
    if (not HasAttribute(game,"aslcalling")) {
      game.aslcalling = false
    }
    if (not game.aslcalling) {
      RunTurnScripts
    }
    if (game.aslcalling) {
      game.aslcalling = false
    }
    UpdateStatusAttributes
    CheckDarkness
    UpdateObjectLinks
  </function>
```

...and here is the new AslCall() function:

```xml
  <function name="AslCall" parameters="data">
    game.aslcalling = true
    stuff = Split(data,"=")
    obj_attr = stuff[0]
    obj_attr = Split(obj_attr,".")
    obj = obj_attr[0]
    obj = GetObject(obj)
    attr = obj_attr[1]
    val = stuff[1]
    if (EndsWith(val,"_toInt")) {
      val = ToInt(Replace(val,"_toInt",""))
    }
    else if (EndsWith(val,"_toDouble")) {
      val = ToDouble(Replace(val,"_toDouble",""))
    }
    set (obj, attr, val)
  </function>
```

---
Here is an example, in which I use the new stuff:
```
JS.eval ("ASLEvent('AslCall','player.aslcall_aka=KV||player.aslcall_int=42_toInt||player.aslcall_double=1.21_toDouble');")
```

---
It sets everything correctly, and doesn't finish the turn.

You can invoke numerous AslCalls in the desktop version, but only one during each turn when playing online, hence the ```||``` split.

</details>

---
I am closing this pull request.  If you'd like, I can create a new one with this new code.

Best regards,
KV

ThePix:
Having thought about it, I think an even better way is to just change the FinishTurn function:

 if (GetBoolean(game, "preventturnscripts")) {
  game.preventturnscripts = false
}
else {
  RunTurnScripts
}
UpdateStatusAttributes
CheckDarkness
UpdateObjectLinks

Then any one can set game.preventturnscripts to true for one "turn" for anything. For example, you might want to do it for the HELP command.

I then do not think we need ASLCall, authors can disable the turnscripts in their callback function if required.

I have not tried online. Why can you only do it your way once?





________________________________

KVonGit:
You're doing pretty much the same thing I ended up with, except your IF...THEN... is set up better than mine.

...and you're not checking for the attribute first (of course, you'll be adding that attribute into the source code, so there will be no need to check).

Groovy!

I just changed mine to match yours (except I still set the attribute to False if it does not exist).

---
> I have not tried online. Why can you only do it your way once?

I sure wish I knew!

It's just one of those things that mysteriously works differently online (like the way it changes the ```.``` to a ```,``` in a double when adding it on to an existing string).

That's why I have ```AslCall()``` set up to break one parameter down into however many I want:

```
  <function name="AslCall" parameters="data">
    game.preventturnscripts = true
    data = Split(data,"||")
    foreach (bit, data) {
      stuff = Split(bit,"=")
      obj_attr = stuff[0]
      obj_attr = Split(obj_attr,".")
      obj = obj_attr[0]
      obj = GetObject(obj)
      attr = obj_attr[1]
      val = stuff[1]
      if (EndsWith(val,"_toInt")) {
        val = ToInt(Replace(val,"_toInt",""))
      }
      else if (EndsWith(val,"_toDouble")) {
        val = ToDouble(Replace(val,"_toDouble",""))
      }
      set (obj, attr, val)
    }
  </function>
```

```
JS.eval ("ASLEvent('AslCall','player.aslcall_aka=KV||player.aslcall_int=42_toInt||player.aslcall_double=1.21_toDouble');")
```

ThePix:
Yes, it is basically what you did, just rearranged.


I used GetBoolean so I do not need the attribute to exist. It returns false if the attribute is set to false or if it does not exist.


________________________________

PR #934, opened 2018-02-06
Add "defaultroom" type to rooms in the editor

KVonGit:
This is just to make it easier to identify rooms.


Pertex:
What is the purpose of this implementation?  In the editor you can check "editor_room" directly and in a published game "editor_room" is removed and so "defaultroom", too.

KVonGit:
Curses!!!

You're right (of course).

I was thinking "editor_room" was removed when you pressed 'Play'.  I forgot it was after publishing.

I don't want to search for objects with descriptions or exits to get a list of all the rooms.

I thought I figured out how to make the editor do the work for me...

---
Removing this request and shedding a single tear.

KVonGit:
@Pertex 

...and...

@ThePix 

I figured out how to do this and make it stick.

Is this a thing anyone besides me would want added?  

If not, no need to read on.  

If yes, let me know, and I'll create a pull request.

Here are the details:

---

I changed "editor_room" to "defaultroom" in EditorController.cs:

Line 1326
```cs
        public void CreateNewRoom(string name, string parent, string alias)
        {
            m_worldModel.UndoLogger.StartTransaction(string.Format("Create room '{0}'", name));
            CreateNewObject(name, parent, "defaultroom", alias);
            m_worldModel.UndoLogger.EndTransaction();
        }
```

---
Then, I added the type "defaultroom" to CoreTypes.aslx.  (This inherits "editor_room".)

---
Finally, I changed all the templates so the ```room``` object inherits "defaultroom" instead of "editor_room".

```
<object name="room">
    <inherit name="defaultroom" />
```

---
Now any new game creates the rooms as type "defaultroom", with the type "editor_room" inherited (so the editor still works correctly).

Also, any new rooms created in existing games inherit "defaultroom" (which inherits "editor_room").

The only issue I see is with existing games.  Any rooms that already exist will not inherit "defaultroom".

The only problem with that is that the function to find all the rooms won't find those rooms, since they already existed and don't have the type the script would filter, which, of course, would be "defaultroom".

Of course, existing games won't be using the function to search for objects with the "defaultroom" type anyway.  So I think it would be okay.

Thoughts, anyone?

Pertex:
Uhhh, a cold shiver runs down my back. :-) Normally I stick to "Never change a running system" and that's a serious change. I don't see any mistakes at the moment, but the devil is in the details. 
I see the need for this implementation, but I would like to see a more comprehensive solution. I would like to have an attribute at each object, e. g. objecttype=room, objecttype=exit, objecttype=male.... so you can not only check rooms, but all objects.
Another idea: Is it possible to add an attribute directly under the new object in EditorController. cs? Something like 
 newObject. fields[FieldDefinitions. ObjectType] = "room";

KVonGit:
> Normally I stick to "Never change a running system" and that's a serious change. 

I would normally stick to the same rule, but quite a few people have wanted an easy way to identify a room over the past few months (myself included).

---
> I don't see any mistakes at the moment, but the devil is in the details.

I say this about everything, too.  (Because it's always true!  Ha-ha!)

---
>I see the need for this implementation, but I would like to see a more comprehensive solution. I would like to have an attribute at each object, e. g. objecttype=room, objecttype=exit, objecttype=male.... so you can not only check rooms, but all objects.

Exits inherit ```defaultexit```.  

Males inherit ```male``` (and/or ```namedmale```).

Containers inherit ```container_base```.

I think the rooms are the only 'odd men out'.

---
> Is it possible to add an attribute directly under the new object in EditorController. cs? 

I tried and tried to do that before I thought of adding the "defaultroom" type.

(I also just tried your suggestion, and I couldn't get it to work.  I tried changing it around a little, too, but I failed miserably.)

---
In the end, I think the type might be better than the attribute (but I'm open to suggestion).

The only issue I've run into is trying to edit a game in a version of Quest without that type included.  Published games run fine, even on 5.6.

The same thing happened with the addition of CoreWearables.  Any games I create in 5.7 can't be edited in any version of Quest which is older than that.

---
Final thoughts:

If you guys don't wish to add this change to code, that won't hurt my feelings at all.

Quest's code is quite intricate, after all, and there could be (and probably are) issues that I've overlooked.

KVonGit:
Finally figured out how to add an attribute in EditorController.cs:


Starts at line 1418:
```
        private void CreateNewObject(string name, string parent, string editorType, string alias)
        {
            Element newObject = m_worldModel.GetElementFactory(ElementType.Object).Create(name);
            if (parent != null)
            {
                newObject.Parent = m_worldModel.Elements.Get(ElementType.Object, parent);
            }
            if (!string.IsNullOrEmpty(alias))
            {
                newObject.Fields[FieldDefinitions.Alias] = alias;
            }
            if (m_worldModel.Elements.ContainsKey(ElementType.ObjectType, editorType))
            {
                newObject.Fields.AddTypeUndoable(m_worldModel.Elements.Get(ElementType.ObjectType, editorType));
            }
            if (editorType == "editor_room")
            {
                newObject.Fields.Set("isroom", true);
            }
        }
```

Pertex:
Fine! In the editor you can change the type of the room to object and vice versa. Does this also change the attribute?

KVonGit:
> In the editor you can change the type of the room to object and vice versa. Does this also change the attribute?

Nope.  I just noticed that last night.

After extensive testing, I have come to the conclusion that the only efficient way to have Quest automatically handle this properly is with types.

...which I can add to my own games.  I don't need this stuff in the source code.

---
I learn by trial and error, you see, and, from this, I have learned to refrain from creating pulling requests unless I'm directly asked to do so.  (This statement is not meant to be taken unpleasantly.  It's all good.)

PR #945, opened 2018-04-08
Adding and modifying multiple things

KVonGit:
Added to playercore.js

```isMobilePlayer()```

Returns a boolean.

```showPopup(title, text)```

Creates and prints a dialog element (like a popup)

```showPopupCustomSize(title, text, width, height)```

Creates and prints a dialog element (like a popup) with custom height and width.

```showPopupFullscreen(title, text)```

Creates and prints a full screen dialog element (like a popup)

**Transcript functions:**

From line 1375 to 1426

```showTranscript()```

Displays the transcript in a dialog element.

```printTranscriptDiv()```

Sends the text in the transcript dialog element to the printer.

---
Changes to CoreCommands:

Modified ```lookat``` to increment the new attribute ```timesexamined``` (when looking is successful)

Modified ```take``` to exclude objects with ```not_all``` set to true

Modified ```version_cmd``` (now it lists more details)

Added ```view_transcript_cmd```

---
Changes to CoreEditor:

Added the ```isroom``` attribute to the **editor_room** type.

---
Changes to CoreEditorObjectInventory:

Added a checkbox to toggle ```not_all```  (This does not effect ```npc_type``` settings!)

image

---
Changes to CoreEditorObjectSetup:

Added control checkbox to toggle ```isroom``` (just in case the object's room settings are changed after object creation)

image

---
Added to CoreFunctions:

```AllRooms()```

Returns an object list of all objects with ```isroom``` set to true.

```DictionaryAdd()```

Adds an item to a dictionary if the dictionary exists.  If the dictionary does not exist, a Log entry is added (an error is not thrown!).

If the dictionary already contains the key, the value will be overwritten!

```DictionaryRemove()```

If a dictionary exists and the key exists, this will remove the key.  (No errors will be thrown!)

```DbgLog()```

Adds log entry if game.debugging is set to true.

```DbgMsg()```

Prints a message if game.debugging is set to true.

---
Change to CoreTypes:

Added ```timesexamined``` integer attribute on **defaultobject**, set to 0

---
EVERY LANGUAGE TEMPLATE:

Added ```isroom``` attribute (set to true) on the **room** object.

---
CoreEditorObjectInventory:

Corrected a typo in my last commit (all included in this pull request, methinks).

---
CoreTypes:

Added ```not_all``` attribute set to false on **defaultobject**.

---
NOTE: 

For some reason, grid.js and playercore.js are updated after my first build when I fork Quest. It looks like the grid-clicking functions get moved from file to file, but I don't mess with either one.  (Everything still seems to work as far as grid-clicking is concerned, too.)

---
Core.aslx:

Line 54

Changed ```JS.SetCompass...``` to ```JS.setCompass...```

---
This passes all 87 tests, and test game produces a valid save file.

---
# NOTE CONCERNING JS DIALOG FUNCTIONS (including log and transcript)

I have no way to test these on a Mac, iPhone, or iPad.

Also, might it be a good idea to include an attribute to toggle whether or not the log can be accessed during play?

### DIFF
--- a/PlayerController/playercore.js
+++ b/PlayerController/playercore.js
@@ -745,14 +745,32 @@ function disableAllCommandLinks() {
     });
 }
 
+// Modified by KV to handle the transcript
+var transcriptEnabled = true;
+var clearedOnce = false;
 function clearScreen() {
-    $("#divOutput").css("min-height", 0);
-    $("#divOutput").html("");
-    createNewDiv("left");
-    beginningOfCurrentTurnScrollPosition = 0;
-    setTimeout(function () {
-        $("html,body").scrollTop(0);
-    }, 100);
+    if (!transcriptEnabled) {
+        $("#divOutput").css("min-height", 0);
+        $("#divOutput").html("");
+        createNewDiv("left");
+        beginningOfCurrentTurnScrollPosition = 0;
+        setTimeout(function () {
+            $("html,body").scrollTop(0);
+        }, 100);
+    } else {
+        $("#divOutput").append("<hr class='clearedAbove' />");
+		  if (!clearedOnce) {
+			addText('<style>#divOutput > .clearedScreen { display: none; }</style>');
+        }
+        clearedOnce = true;
+        $('#divOutput').children().addClass('clearedScreen');
+        $('#divOutput').css('min-height', 0);
+        createNewDiv('left');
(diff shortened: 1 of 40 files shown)


KVonGit:
I have added a few things to this, so I'm closing this pull request and creating a new one which includes everything.

PR #946, opened 2018-04-11
Adding and modifying multiple things (version 2)

KVonGit:
Added to playercore.js

```isMobilePlayer()```

Returns a boolean.

```showPopup(title, text)```

Creates and prints a dialog element (like a popup)

```showPopupCustomSize(title, text, width, height)```

Creates and prints a dialog element (like a popup) with custom height and width.

```showPopupFullscreen(title, text)```

Creates and prints a full screen dialog element (like a popup)

**Transcript functions:**

From line 1375 to 1426

```showTranscript()```

Displays the transcript in a dialog element. (This can be disabled by calling ```DisableTranscript``` in Quest.)

```printTranscriptDiv()```

Sends the text in the transcript dialog element to the printer.

```clearScreen()``` has been modified to handle the transcript.

If the JS variable ```enableTranscript()``` is set to false, the old ```clearScreen()``` script will run.

```enableTranscript()``` is set to true by default.  If the Quest attribute game.notranscript is changed to false, the default ```StartGame``` function will set the JS variable ```enableTranscript()``` to false.  

(I also added a ```DisableTranscript``` function to Quest to handle this.  Just add it to the start script.)

---
Changes to Core:

Modified StartGame to set the JS variable ```enableTranscript()``` to false if game.notranscript is set to true.  (This will run the old JS ```clearScreen()``` script.)

---
Changes to CoreCommands:

Modified ```lookat``` to increment the new attribute ```timesexamined``` (when looking is successful)

Modified ```take``` to exclude objects with ```not_all``` set to true

Modified ```version_cmd``` (now it lists more details)

Added ```view_transcript_cmd```  (This can be disabled by adding ```DisableTranscript``` to your start script or by setting ```game.notranscript``` to true.)

Added IF statement to ```log_cmd``` which will inform the player the log is not enabled if you add ```DIsableHtmlLog``` to your start script OR if you change ```game.nohtmllog``` is true.

---
Changes to CoreEditor:

Added the ```isroom``` attribute to the **editor_room** type.

---
Changes to CoreEditorObjectInventory:

Added a checkbox to toggle ```not_all```  (This does not effect ```npc_type``` settings!)

image

---
Changes to CoreEditorObjectSetup:

Added control checkbox to toggle ```isroom``` (just in case the object's room settings are changed after object creation)

image

---
Added to CoreFunctions:

```AllRooms()```

Returns an object list of all objects with ```isroom``` set to true.

```DictionaryAdd()```

Adds an item to a dictionary if the dictionary exists.  If the dictionary does not exist, a Log entry is added (an error is not thrown!).

If the dictionary already contains the key, the value will be overwritten!

```DictionaryRemove()```

If a dictionary exists and the key exists, this will remove the key.  (No errors will be thrown!)

```DbgLog()```

Adds log entry if game.debugging is set to true.

```DbgMsg()```

Prints a message if game.debugging is set to true.

```DisableTranscript()```

Prevents the player from opening the transcript dialog popup and causes the JS ```clearScreen()``` function to run the old script (which removes the cleared elements rather than hiding them).  This prints a message saying the game has no transcript feature.

```DisableHtmlLog()```

Prevents the player from viewing the in-game log dialog popup.  This prints a message saying the game has no in-game log.

---
Change to CoreTypes:

Added ```timesexamined``` integer attribute on **defaultobject**, set to 0

Added ```nohtmllog``` and ```notranscript``` attributes to the game object, both set to False.

---
EVERY LANGUAGE TEMPLATE:

Added ```isroom``` attribute (set to true) on the **room** object.

Also added ```log_cmd```, ```version``` and ```view_transcript_cmd``` templates to every language except Greek* (these are in English).

* The Greek template has a pending pull request.  If I include my changes to Greek, it creates a merging issue.

---
CoreEditorObjectInventory:

Corrected a typo in my last commit (all included in this pull request, methinks).

---
CoreTypes:

Added ```not_all``` attribute set to false on **defaultobject**.

---
NOTE: 

For some reason, grid.js and playercore.js are updated after my first build when I fork Quest. It looks like the grid-clicking functions get moved from file to file, but I don't mess with either one.  (Everything still seems to work as far as grid-clicking is concerned, too.)

---
Core.aslx:

Line 54

Changed ```JS.SetCompass...``` to ```JS.setCompass...```

---
This passes all 87 tests, and a test game produces a valid save file.

---
# NOTE CONCERNING JS DIALOG FUNCTIONS (including log and transcript)

I have no way to test these on a Mac, iPhone, or iPad.

### DIFF
--- a/PlayerController/playercore.js
+++ b/PlayerController/playercore.js
@@ -745,14 +745,32 @@ function disableAllCommandLinks() {
     });
 }
 
+// Modified by KV to handle the transcript
+var transcriptEnabled = true;
+var clearedOnce = false;
 function clearScreen() {
-    $("#divOutput").css("min-height", 0);
-    $("#divOutput").html("");
-    createNewDiv("left");
-    beginningOfCurrentTurnScrollPosition = 0;
-    setTimeout(function () {
-        $("html,body").scrollTop(0);
-    }, 100);
+    if (!transcriptEnabled) {
+        $("#divOutput").css("min-height", 0);
+        $("#divOutput").html("");
+        createNewDiv("left");
+        beginningOfCurrentTurnScrollPosition = 0;
+        setTimeout(function () {
+            $("html,body").scrollTop(0);
+        }, 100);
+    } else {
+        $("#divOutput").append("<hr class='clearedAbove' />");
+		  if (!clearedOnce) {
+			addText('<style>#divOutput > .clearedScreen { display: none; }</style>');
+        }
+        clearedOnce = true;
+        $('#divOutput').children().addClass('clearedScreen');
+        $('#divOutput').css('min-height', 0);
+        createNewDiv('left');
+        beginningOfCurrentTurnScrollPosition = 0;
+        setTimeout(function () {
+            $('html,body').scrollTop(0);
+        }, 100);
+    }
 }
 
 function keyPressCode(e) {
@@ -1405,9 +1423,11 @@ function showTranscript() {
         modal: true,
     });
     $('#transcriptdata').html($('#divOutput').html());
-    $("#transcriptdata *").attr('style', '').attr('color', '');
-    transcriptDialog.dialog("open");
-
+    $("#transcriptdata a").addClass("disabled");
+	transcriptDialog.dialog("open");
+	setTimeout(function () {
+	    $("#transcriptdata a").addClass("disabled");
+	}, 1);
 };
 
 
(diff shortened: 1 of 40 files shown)


ThePix:
For the other languages, it would be better to omit the templates, rather than including them in English. The language files include English first, so any missing templates will appear in English anyway, and it is easier to work out what needs doing if the template is missing, rather than in English.

KVonGit:
I removed the templates from every file except English.aslx

PR #947, opened 2018-04-14
Update CoreDescriptions.aslx

KVonGit:
Modified ```GetListDisplayAlias``` (again)

Line 376:  Now sends the alias through ```ProcessText()```, even if there is no ```listalias``` attribute.

Allows us to use an alias like this:
```book ({either book.isopen:open|closed})```

image

### DIFF
--- a/WorldModel/WorldModel/Core/CoreDescriptions.aslx
+++ b/WorldModel/WorldModel/Core/CoreDescriptions.aslx
@@ -368,12 +368,12 @@
 
   <function name="GetListDisplayAlias" type="string" parameters="obj">
     <![CDATA[
+    // Modified by KV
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ThePix** [comment]
Can you describe when you would use this. The alternative approach is to set the listalias when the book is open or closed, and I am not seeing why your way makes it any simpler.

**Comment by KVonGit** [comment]
I would use it any time using the text processor to control what the player sees as an object's alias.

...BUT this is dangerous, and I meant to remove this request shortly after I created it.

If someone were to use ```{either book.isopen:open|closed} book```, that would cause parsing problems.

```
You are in a room.
You can see a closed book.

> x closed
I can't see that.


> x book
Nothing out of the ordinary.
```

PR #948, opened 2018-04-20
Update Menu.resx

KVonGit:
Lines 146 - 148

Added ```Ctrl+R``` shortcut keys data. (Without this, ```CTRL+R``` does nothing at all.)

---
I've found an easy way to add a RESTART command, as well.

https://github.com/KVonGit/QuestStuff/wiki/Adding-a-RESTART-command-(requires-VS-to-build-from-source-code)

Let me know if you'd like a pull request with that stuff.

### DIFF
--- a/Menus/Menu.resx
+++ b/Menus/Menu.resx
@@ -143,6 +143,9 @@
   <data name="RestartToolStripMenuItem.ImageScaling" type="System.Windows.Forms.ToolStripItemImageScaling, System.Windows.Forms">
     <value>None</value>
   </data>
+  <data name="RestartToolStripMenuItem.ShortcutKeys" type="System.Windows.Forms.Keys, System.Windows.Forms">
+    <value>Ctrl+R</value>
+  </data>
   <data name="RestartToolStripMenuItem.Size" type="System.Drawing.Size, System.Drawing">
     <value>257, 26</value>
   </data>


ThePix:
Would this (RESTART command) work on the web player? I ask because it starts in a new window.

KVonGit:
Yes.

Your key is "Strg".

I just researched this, and it looks like we can replace "Ctrl" with "Control" to make it work for everyone.

https://social.msdn.microsoft.com/Forums/vstudio/en-US/9ab50ba0-9be1-49e8-b4f6-7cfd97eccf10/msbuild-cannot-process-resource-files?forum=msbuild


---
Should I do a pull request, or do you think it would be better if you tried replacing ```Ctrl+R``` with ```Control+R``` in your Menu.resx file to see if it works first?

Pertex:
Are these shortcuts all new?

KVonGit:
No, sir.

I'm in that file now, and I see Ctrl+N, Ctrl+O, and 5 or 6 more.

I only added Ctrl+R, I promise.

If that information was correct, changing each Ctrl to Control should fix it for everyone.  I'm trying it now to see if it breaks it in English.

KVonGit:
Well, they may be new to you if you haven't done this since Pixie updated to 5.8.

I don't know if the old Menu.resx had 'Ctrl' shortcuts or not, but I know Ctrl+R worked before updating to 5.8, so some sort of code existed.  If this fix doesn't work, I'll look at the old file.

Pertex:
Control seems to work here, you could change it

KVonGit:
I will edit the file and submit a new pull request.

Sorry about that!

Pertex:
I found CTRLs in three files: Menu.resx, WFEditorTree.de-DE.resx and WFEditorTree.resx

KVonGit:
I changed the ones in Menu.resx, built Quest, and made sure the Ctrl shortcuts still work.  They do.

I assume the other changes will work, too, but I can't test them before submitting a pull request.

Have you tested out the shortcut keys in a fresh build, Pertex?  (Just making sure before I change things I can't test.)

KVonGit:
I tested it all in English, and it works fine.  The shortcut keys work in the build, as well.

I have submitted 3 pull requests.

Pertex:
fresh build? I have cloned the rep and tried to run Quest in the editor. After changing the value to CONTROL I can start Quest and it seems to work correctly, shortcuts are working

PR #959, opened 2018-05-01
CoreDevMode

SoonGames:
Customized dev mode, including custom templates in EditorGerman.aslx and EditorEnglish.aslx. Where the DevMode is exactly integrated I leave Pixie. The function for calling is 'setDevMode'.


KVonGit:
The templates in EditorEnglish.aslx are keeping gamebooks from opening in the editor.

See this post:

https://github.com/textadventures/quest/issues/969#issuecomment-386952829

SoonGames:
Oh sorry. I must have overlooked that here. Next time, though.

PR #966, opened 2018-05-03
Splitter in attribute tab on half of window size

SoonGames:
Set the splitter in the attribute tab to half the window size to make the assignments easier to use.


SoonGames:
preview2

PR #980, opened 2018-05-08
Distribution of the German templates for the DevMode

SoonGames:


SoonGames:
Why conflict? I understand nothing.

KVonGit:
> Why conflict?

If you have GitHub desktop, it sometimes gives you a hint.

I've never seen it throw a conflict error unless I was committing a file which was in an open pull request, and I don't see that happening.

So, if it makes you feel better, I'm just as confused by this as you are.

PR #982, opened 2018-05-10
Change to v580

KVonGit:
Due to recent changes in the C# code, previous versions of Quest will not be able to handle newly created games.  The ASL version change will prevent previous versions of Quest from attempting to open these files.

---
Please, feel free to close this without merging if this is not the route you'd like to take.


ThePix:
Given the changes to the underlying code, I agree this is required.

PR #983, opened 2018-05-11
Change to v580

KVonGit:
I missed this one.


ThePix:
There is a principle in coding called DRY - Don't Repeat Yourself. The idea is that if you need to subsequently make a change, then you just need to do it in one place, not over a dozen across several files. The developers of Visual Studio seem to have not heard of it!

PR #985, opened 2018-05-11
Update ReturnScript.cs

KVonGit:
Changed v580 back to v550 on line 49, so version 550 ASL games will still function correctly when played from the editor.


KVonGit:
How does this tie in with the web editor?  (Will the web editor use this file?  If not, I don't see how the ASL version effects the web editor.)

PR #1000, opened 2018-05-23
Update CoreCommands.aslx

KVonGit:
Changed this bit in the "drop" command from ```NothingToTake``` to ```NotCarryingAnything```:

```c
      if (multiple and ListCount(object) = 0) {
        msg("[NotCarryingAnything]")
      }
```


ThePix:
Good spot!

There is a template for this, NothingToDrop, but as we only have it in six languages, and NotCarryingAnything in 16, we will go with your suggestion.

PR #1001, opened 2018-05-23
Update playercore.js

KVonGit:
```clearScreen()```  improved by mrangel.

Now we can append elements with the class 'donotclear' to divOutput and ```clearScreen()``` will ignore those elements.

http://textadventures.co.uk/forum/quest/topic/n-jjyktjbeg9vqwnsyy0kw/stackable-transparent-images-preloading-stuff#86e9b665-cd0d-4bcf-8898-b3cfbfae7159

### DIFF
--- a/PlayerController/playercore.js
+++ b/PlayerController/playercore.js
@@ -761,29 +761,27 @@ function disableAllCommandLinks() {
 // Modified by KV to handle the scrollback feature
 var saveClearedText = true;
 var clearedOnce = false;
+/* 
+ * Improved by mrangel
+ * Now we can append elements with the class 'donotclear' to divOutput and clearScreen() will ignore those elements.
+*/
 function clearScreen() {
     if (!saveClearedText) {
-        $("#divOutput").css("min-height", 0);
-        $("#divOutput").html("");
-        createNewDiv("left");
-        beginningOfCurrentTurnScrollPosition = 0;
-        setTimeout(function () {
-            $("html,body").scrollTop(0);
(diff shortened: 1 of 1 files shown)


ThePix:
How well tested is this?

Would we need to change webplayer too?

KVonGit:
I've tested it with all sorts of games that clear the screen, but only in the desktop player.

---
I'm 95% sure **WorldModel\WorldModel\playercore.js** gets copied to **WebPlayer\playercore.js**.  If so, the change should effect everything.  This is a theory.  I don't see ```clearScreen()``` in any other JS file, so the one in playercore.js should be king.

KVonGit:
It works well online when I include it in a JS file.  I can't really run the web player from VS, because my computer can't keep up.

Here's my test game.

TEST to clear the screen (printing "This is a test." before clearing).

LOOK to show the room description.

(I was not very creative.)

http://textadventures.co.uk/games/view/mffwrogv0eyn7dkfhlydmg/new-clearscreen-tester

ThePix:
> I'm 95% sure WorldModel\WorldModel\playercore.js gets copied to WebPlayer\playercore.js.

I have now confirmed that, so that is not an issue.

KVonGit:
This needs improvement.

Start reading from this point:

http://textadventures.co.uk/forum/quest/topic/hwvergkpgeooythoggn3jq/this-should-work-but-solved#1d5d40ea-bf1a-4de1-8cca-342193232073

PR #1010, opened 2018-05-27
Update CoreOutput.aslx

KVonGit:
Modified Log to use JSSafe to avoid browser errors with the new HTML log feature.

### DIFF
--- a/WorldModel/WorldModel/Core/CoreOutput.aslx
+++ b/WorldModel/WorldModel/Core/CoreOutput.aslx
@@ -849,10 +849,11 @@
   <function name="Log" parameters="text">
     <![CDATA[
     //request (Log, text)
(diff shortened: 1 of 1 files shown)


KVonGit:
This is not good because it removes spaces.

I am closing this one.

I also notice the call to the old log has been commented out.  Did I do that?

---
Must find a way to pass ```'``` in the parameter.  That may be the only offensive character...

KVonGit:
This seems to be a good way to handle it:

```c#
// request (Log, text)
txt = Replace(text, "'", "")
if (not GetBoolean(game, "nohtmllog")) {
  JS.eval ("if(typeof(addLogEntry)===\"function\"){ addLogEntry('" + txt + "'); };")
} 
JS.eval ("if(!webPlayer && typeof(WriteToLog)===\"function\"){var s = '" + txt + "';WriteToLog(s);}")
```

PR #1013, opened 2018-06-01
Fix transcript_off and Remove game.notarealturn

KVonGit:
Corrected incorrect variable name in transcript_off_cmd which prevented the transcript from being disabled.

---
While I was in here, I removed all of the lines of code where I declared game.notarealturn, since that was an unused attribute.


KVonGit:
I've edited the first commit and messed this up.  Closing and starting a new one.

PR #1020, opened 2018-06-04
Error fixed when DevMode is switched off. Multicommands separated by a dot and the word "and"

SoonGames:


SoonGames:
The Updates section can be ignored. Sorry. I'm a GitHub beginner. ^^

SoonGames:
Shit. I hadn't thought of that. Thanks for the info.

KVonGit:
Maybe a ```multiple_command_separator``` attribute on the game object would be good?  Then, the author could easily change it, and that line in the script could just be:

 ```Replace (command, game.multiple_command_separator, ".")```

---
Or make the attribute a string list!

```c
if (ListCount (game, "multiple_command_separator) > 0) {
  foreach (s, game.multiple_command_separator){
    Replace (command, s, ".")
  }
}
```

PR #1033, opened 2018-06-22
Object is a Room-Template created

SoonGames:


ThePix:
Sorry about that one two. It was KV's code, and he isn't (or wasn't) up to speed on internationalisation, and I should have spotted it.

SoonGames:
no problem.

PR #1044, opened 2018-08-21
Update Blank.htm

KVonGit:
Fix to load the ```.js``` and ```.css``` files from the DevMode directory


ThePix:
I have used this as the basis of a possible solution, as further files need changing.

PR #1078, opened 2019-06-17
DevMode 2.5.1

scriptgames:
Updated DevMode to latest version.


KVonGit:
@8ne 

Hello.

I've got DevMode fully functional in the testing version of the WebPlayer app, and I must say: it is nice!

Due to recent changes to some of the same files, this PR will have to be closed and then slightly modified to avoid conflicts.

If you are still active, I expect you might want to be involved. (If not, I can make the changes, which won't effect functionality or anything noticeable by the user.)

I will post another comment here when all the new changes have been merged, after it's safe to submit new pull requests.

Thanks for all you've done for the Quest community!

PR #1186, opened 2024-08-26
Fix major bug concerning transcript

KVonGit:
This removes the majority of the code related to the transcript.

Currently, enabling a transcript crashes online games, and it stops all turns scripts from firing while it is enabled in the desktop player.

There is a totally separate, existing function in playercore.js named `showScrollback()` which handles displaying and printing the transcript all by itself. All I left intact besides that JS function are the commands and templates to view the transcript.

I altered docs\transcript.md to reflect these changes.

I also removed or disabled all the Quest and JS functions to save the transcript to a local file in the desktop version, because this slows the games down tremendously in practice.

I left the code in Player\PlayerHTML.vb to handle it if someone wanted to add the code to Quest and JS to use this feature. That way, no one would have to rebuild the project in Visual Studio just to use that functionality.

---
Double check me on the Dansk `transcript_on_cmd` change I made. The internet says it translates correctly everywhere I looked it up.

### DIFF
--- a/Player/PlayerHTML.vb
+++ b/Player/PlayerHTML.vb
@@ -126,6 +126,9 @@ Public Class PlayerHTML
         My.Computer.FileSystem.WriteAllText(logPath + "\" + gameName + "-log.txt", data + Environment.NewLine, True)
     End Sub
     Private Sub SaveTranscript(data As String)
+        ' NOTE FROM KV:
+        '   I disabled or removed all the Quest and JS functions concerning this, but I will leave this code in here.
+        '   This way, someone could simply add the Quest and JS functions to their game if they wished to use this functionality.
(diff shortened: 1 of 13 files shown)


KVonGit:
Perhaps I was a bit rash when I decided to gut most of the transcript stuff.

I am currently testing a build with less changes. Basically, you would have to set `game.writetranscripttofile` to `true` for the game to even call `SaveTranscript()`. And I've added the following to **WebPlayer\player.js** just as a fallback:

```js
function SaveTranscript(data) {
  console.warn("[QUEST] Ignoring call to SaveTranscript(). Cannot write directly to local file when playing online.");
}
```

That fixes the issues with the webplayer.

---
Adding `AslCall()` to the Quest code and `ASLCall()` to **Player\desktopplayer.js** fixes the Turn Script and Walkthrough issues in the desktop player.

---
I shall be testing thoroughly for at least two days before submitting an Option B.

PR #1187, opened 2024-09-02
Update source_code.md

KVonGit:
Fixing the links to load the game in your browser when running the web player or web editor from VS. (Changed 1 to 0.)


KVonGit:
I recently learned it doesn't matter if you use `http://localhost:52426/Play.aspx?id=0` or `http://localhost:52426/Play.aspx?id=1` or `http://localhost:52426/Play.aspx?id=42` when using the debug plugins.

KVonGit:
I ***recently*** learned that the ID is used for saving/resuming purposes, among other things.

Once upon a time, we did need to use `1` for the ID for the saving progress using the debug plugins, but the code to do that no longer exists.

Point is: the ID matters when you can save and resume progress.

PR #1190, opened 2024-09-04
kv-3-log-rehab (Log Redux (Bug Fix))

KVonGit:
The old log is still in Quest 5.8. There is even still a button for it, but it just opens a blank window.

It was left that way because the plan was to deprecate the `request` script, but that ended up being a no-go due to backwards compatibility.

---
- Reinstated the old log functionality

- Replaced concatenated string in JS with an array.

- Replaced the jQuery-UI popup (which messed up walkthroughs) with a normal HTML element, just beneath the command input element, with buttons to hide or print the log.

- Added Quest function `ClearHtmlLog()` to empty the array and the HTML and hide the log div.

- `game.nohtmllog` is set to `true` by default. Change it to `false` to enable the in-game log.

- `game.writelogtofile` is set to `false` by default. Change it to true to write/append directly to a file in "Documents\Quest Logs" when using the desktop player, which comes in very handy if a game freezes or crashes!

- `game.useconsolelog` is set to `false` by default. Change it to `true` to have the log send messages to the HTML dev tools console log.

### DIFF
--- a/PlayerController/playercore.js
+++ b/PlayerController/playercore.js
@@ -357,15 +357,23 @@ function SetBackgroundOpacity(opacity) {
 }
 
 function setBackground(col) {
+    /* If '#rgb', convert to '#rrggbb'*/
+    /* https://github.com/textadventures/quest/issues/1052 */
+    if (col.charAt(0) === "#") && col.length == 4) {
+        var colBak = "" + col + "";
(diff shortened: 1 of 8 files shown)


KVonGit:
I missed a spot. Luckily, Pertex pointed it out before it was too late.

PR #1192, opened 2024-09-07
kv-4-transcript-repair

### LINKED ISSUE #1054: Transcript crashes games online
Enabling a transcript online crashes (freezes) the game.

I seem to remember writing code to handle online transcripts in a dialog popup, but it was long ago and I'm not positive my recollection is correct.

KVonGit:
There is a game-crashing bug in the transcript code I submitted for Quest 5.8.0.

If playing online and the transcript is enabled, the game completely crashes (because `SaveTranscript` is only in desktop.js, and is therefore undefined in the web player).

If playing in the desktop and the transcript is enabled, all turn scripts stop working completely and if a walkthrough is being recorded, the recorded steps are the HTML being output to the game.

ALSO, there was a JS string variable named `transcriptString` which got larger and larger when adding the current text for the transcript every time text printed to the game. This obviously slowed things down. Plus, this text was also sent by `ASLEvent` to update `game.transcriptstring` each time text printed along with everything else.

Also, the updated `clearScreen()` function in playercore.js was set up to save the cleared text and hide it rather than deleting it. This also negatively effected game performance exponentially as play progressed. The JS boolean `saveClearedText` is now set to false by default.

Finally, if the VIEW TRANSCRIPT command was run while recording a walkthrough, the jQuery-UI popup would break the walkthrough during playback.

The in-game HTML transcript feature has now been removed. There is no VIEW TRANSCRIPT command anymore. There is no longer a `transcriptString` JS variable, nor is there a `game.transcriptstring` attribute.

The only way the transcript will work now is if the game is being played in the desktop player. It will save the transcript, and append it, every time text prints to the screen, to an HTML file in "Documents\Quest Transcripts", just as in 5.8.0. Except I also fixed my error in PlayerHTML.vb to make the file save under the transcript name the player chooses when enabling the transcript.

I added a "fallback" of each function I added to Player\desktopplayer.js to WebPlayer\player.js. These functions do nothing and serve no purpose except to avoid possible errors due to being undefined.

Also added a bit of code in `HandleCommand` <s>and `ShowMenuResponse`</s> to add the player's command to the transcript if `game.echocommand` is `false`.

Updated the transcript doc page to reflect the changes.

I triple checked things every way I could imagine in the desktop build as well as using VS to test the web player and the mobile version, but I haven't had anyone else test it. So, test it out, please.

---
I almost forgot:

If `saveClearedText` is changed to `true`, it will save the cleared text in #divOutput rather than dumping it.

The `printScrollback()` JS function is still present. Calling it will use the player's device to print the everything in #divOutput (whether via a printer or saving it to PDF). 

Again, the scroll back will not contain any cleared text unless `saveClearedText` is changed from `false` to `true`.

The scrollback JS functions are pretty much hidden, but an author could easily add code to allow their use in their game.

---
All apologies to everyone for the bugs I introduced in Quest 5.8.0. I believe I have now fixed them all.

---
Closes #1054

### DIFF
--- a/Player/Player.vb
+++ b/Player/Player.vb
@@ -792,7 +792,8 @@ Public Class Player
     End Sub
 
     Private Sub ctlPlayerHtml_SendEvent(eventName As String, param As String) Handles ctlPlayerHtml.SendEvent
-        If RecordWalkthrough IsNot Nothing Then
+        ' Don't save the step if calling WhereAmI or any event starting with NoEventFunc
+        If RecordWalkthrough IsNot Nothing And Not eventName.StartsWith("WhereAmI") And Not eventName.StartsWith("NoEventFunc") Then
             m_recordedWalkthrough.Add("event:" + eventName + ";" + param)
         End If
         m_game.SendEvent(eventName, param)
--- a/Player/PlayerHTML.vb
+++ b/Player/PlayerHTML.vb
@@ -143,15 +143,15 @@ Public Class PlayerHTML
         If Not System.IO.Directory.Exists(transcriptPath) = True Then
             System.IO.Directory.CreateDirectory(transcriptPath)
         End If
-        If Not System.IO.File.Exists(transcriptPath + "\" + mgameName + "-transcript.html") = True Then
+        If Not System.IO.File.Exists(transcriptPath + "\" + mgameName + "-transcript.txt") = True Then
             Dim file As System.IO.FileStream
-            file = System.IO.File.Create(transcriptPath + "\" + mgameName + "-transcript.html")
+            file = System.IO.File.Create(transcriptPath + "\" + mgameName + "-transcript.txt")
             file.Close()
         End If
         If data.Contains("___SCRIPTDATA___") Then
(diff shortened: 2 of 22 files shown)


alexwarren:
To help me get more of the context here, can you point me to the commit(s) / PR(s) where this stuff was originally added?

KVonGit:
> To help me get more of the context here, can you point me to the commit(s) / PR(s) where this stuff was originally added?

Hello.

https://github.com/textadventures/quest/commit/a665d25b8b9e41996a1cb4124c6c34abd7686b9d

---
I'm actually working on the transcript functionality right now.

I recently changed it from saving to an HTML file to a TXT file, and I noticed there were too many new lines being added to the transcript file. I've got that fixed now.

I was about to get my branch straightened out and submit a pull request, when I saw this notification.

I'm deleting this branch (closing this pull request) and submitting a new pull request from a fresh branch. I've made so many changes in this branch while testing, it is a mess.

---
By the way, do you think the transcript should even be included by default?

PR #1198, opened 2024-09-25
kv-7-pertex-location-bar-background-image-option

KVonGit:
Submitted by Pertex

Adds the option to set your own location bar background image in the editor


KVonGit:
This was based on kv-6*, and I just had to revise that pull request. Resubmitting this one shortly.

PR #1235, opened 2024-11-24
Add CSharp SyntaxHighlighting to Script TextBoxes

TextMisadventures:
Would it also be possible to apply the word wrap settings from full code view to script text-boxes?


alexwarren (approved):

alexwarren:
@TextMisadventures I'm happy for the word wrap settings to match if you can submit a PR for that (or add it as an Issue, and I'll get round to having a look at some point)

PR #1239, opened 2024-11-24
Pertex - Logo and Desktop Icon, DateTime, endWaitDiv, Filewatcher Core

### LINKED ISSUE #1228: continue is missing
If you disable the command bar, the continue-link ("Wait for keypress") is not visible.

The commandbar is in div txtCommandDiv and the endWaitlink is in this txtCommandDiv, too. Is it possible to move the endWaitlink into a new div?

[bug.zip](https://github.com/user-attachments/files/17801752/bug.zip)

TextMisadventures:
Submitted by @Pertex 

- Filewatcher watches changes in Core
- endWaitDiv allows displaying endWaitLink with no command bar
  - closes #1228 
- DateTime hard-coded string functions: `CurrentDate` and `CurrentTime`
- Touch-up logo and desktop icon


Editor/Editor.vb
```diff
@@ -65,7 +66,11 @@ Public Class Editor
                             Dim path As String = System.IO.Path.GetDirectoryName(m_filename)
                             Dim filter As String = System.IO.Path.GetFileName(m_filename)
                             m_fileWatcher = New System.IO.FileSystemWatcher(path, "*.aslx")
+                            m_fileWatcher.IncludeSubdirectories = True
                             m_fileWatcher.EnableRaisingEvents = True
+                            m_fileWatcherQuest = New System.IO.FileSystemWatcher(Application.StartupPath + "\Core", "*.aslx")
```
alexwarren, line 71: I'm not entirely convinced of this change, I don't think anyone should be editing core library `.aslx` files directly within the Quest app directory itself, while the editor is loaded?

Editor/Editor.vb
```diff
@@ -421,6 +426,7 @@ Public Class Editor
     Private Function Save(filename As String) As Boolean
         Try
             m_fileWatcher.EnableRaisingEvents = False
+            m_fileWatcherQuest.EnableRaisingEvents = False
```
alexwarren, line 429: Regardless... if `m_fileWatcherQuest` is only monitoring the Quest app directory, there's no need to deactivate it here when a game file is saved

Editor/Editor.vb
```diff
@@ -1021,6 +1028,14 @@ Public Class Editor
         Reload()
     End Sub
 
+    Private Sub m_fileWatcherQuest_Changed(sender As Object, e As System.IO.FileSystemEventArgs) Handles m_fileWatcherQuest.Changed
```
alexwarren, line 1031: This duplicates the regular file watcher, so you could use the same event handler function here (though I would rename it). Although, if you were monitoring the app directory, you would want the banner to say that (right now it will tell you that the game's `.aslx` file has changed, when it hasn't).

WorldModel/WorldModel/Functions/DateTimeFunctions.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Text;
+
+namespace TextAdventures.Quest.Functions
+{
+    public static class DateTimeFunctions
+    {
+        public static string CurrentDate()
+        {
+            DateTime localDate = DateTime.Now; 
+            return localDate.ToShortDateString();
```
alexwarren, line 13: These return locale-specific strings i.e. the format will depend on the locale of the machine. That may be fine for running on desktop, but on the web it will always be the locale of the server.

Also, because you've only provided these locale-specific versions, there's no way for a game to reliably process the date (if it wanted to extract the current month for example - how does it know where in the string that is).

So this feature needs a lot more thought before we can add it.

alexwarren (changes requested):
I like the tweaked logo. I've left comments on a bunch of other changes here.

These kinds of changes would be much easier to review as separate PRs. There are 4 independent changes here, so could you put these in 4 separate PRs please - that way they can be merged and/or discussed separately.

Editor/Editor.vb
KVonGit: This was all @Pertex. (Not sure if he gets notified without tagging him here.)

We never really discussed this one; so, I don't know what his reason was. I'm sure he'll be along soon, though.

Pertex: I have added these things especially for quest developers, when you need to quickly edit core files. An addition, but not important for the normal user.

Editor/Editor.vb
Pertex: My knowledge is not sufficient for this

WorldModel/WorldModel/Functions/DateTimeFunctions.cs
Pertex: That's right, we've also racked our brains over it. We can only get the current local time via the browser and Javascript. And since the communication between Quest and Javascript is not easy, I first created two simple functions that a developer can already do something with. The functions and results are not wrong, you can work with them. I use them, for example, to determine the playing time from the start of a game to the end in days.

Perhaps you have a tip for us on how we can send data from Javascript to Quest?

alexwarren: You would use `ASLEvent` for that?

Pertex: There were concerns about this because there were difficulties with ASLEvent, timing problems I think. Let me discuss this again with KV plz

KVonGit: > These return locale-specific strings i.e. the format will depend on the locale of the machine. That may be fine for running on desktop, but on the web it will always be the locale of the server.

> That's right, we've also racked our brains over it. We can only get the current local time via the browser and Javascript.

That's why I added the UTC counterparts. We didn't realize this at first, and we'd added time and date attributes to the game object in start script. After we compared our online transcripts one day, we noticed the time was that of the server, and I removed the bits from the Core library, added the UTC functions, and left it in the source code so people could use it if they wanted it.

If I were to use this in a game:

<details><summary> Example game </summary>

[Do You Have the Time (For the Quest 5.9 testing build only).zip](https://github.com/user-attachments/files/18263949/Do.You.Have.the.Time.For.the.Quest.5.9.testing.build.only.zip)

```xml
<asl version="580">
  <include ref="Core.aslx" />
  <game name="Do You Have the Time?">
    <gameid>eda4b815-d1ee-4203-b442-d0559f12654d</gameid>
    <version>1.0</version>
    <feature_advancedscripts />
    <attr name="autodescription_youarein_useprefix" type="boolean">false</attr>
    <attr name="autodescription_description" type="int">2</attr>
    <attr name="autodescription_youcango" type="int">4</attr>
    <attr name="autodescription_youcansee" type="int">3</attr>
    <autodescription_description_newline />
    <inituserinterface type="script">
      SetUpInitTimeAndDate
    </inituserinterface>
    <start type="script">
      SetUpStartTimeAndDate
    </start>
  </game>
  <object name="room">
    <isroom />
    <description><![CDATA[<br/>The current time is {=DictionaryItem (GetCurrentTime(),"time")}{if not game.islocaltime: UTC}, and the date is {=DictionaryItem (GetCurrentDate(), "date")}.<br/><br/>JavaScript Date(): <span id="js-date-{game.currentturnoutputsection}"><img src="{=GetFileURL("foo")}" onerror="updateJsDateDiv('{game.currentturnoutputsection}');$(this).remove();"/>loading...</span><br/><br/>You started this game at {game.starttime} on {game.startdate}.<br/><br/>You loaded the current session at {game.inittime} on {game.initdate}.<br/><br/>Wait until the time changes and type {command:LOOK} to test this room description.<br/><br/>Use the {command:SAVE} command, then load the saved game to test the start and init time and date attributes on the game object.<br/><br/>]]></description>
    <alias>Time and Date Testing Grounds</alias>
    <usedefaultprefix type="boolean">false</usedefaultprefix>
    <object name="player">
    </object>
  </object>
  <function name="IsDesktop" type="boolean">
    return (GetFileURL("BaseURI").StartsWith("quest"))
  </function>
  <function name="GetCurrentTime" type="dictionary">
    if (IsDesktop()) {
      return (QuickParams ("time", CurrentTime(),"local", true))
    }
    return (QuickParams ("time", CurrentUtcTime(),"local", false))
  </function>
  <function name="GetCurrentDate" type="dictionary">
    if (IsDesktop()) {
      return (QuickParams ("date", CurrentDate(), "local", true))
    }
    return (QuickParams ("date", CurrentUtcDate(), "local", false))
  </function>
  <function name="SetJsDate" parameters="jsdate"><![CDATA[
    game.changedjsdate => {
      JS.updateJsDateDiv (game.currentturnoutputsection, game.jsdate)
    }
    game.jsdate = jsdate
  ]]></function>
  <function name="SetUpInitTimeAndDate">
    timedict = GetCurrentTime()
    game.inittime = DictionaryItem (timedict, "time")
    game.islocaltime = DictionaryItem (timedict, "local")
    game.initdate = DictionaryItem (GetCurrentDate(), "date")
    JS.eval("window.updateJsDateDiv = function(idsuffix, s=Date().toString()) {$('#js-date-' + idsuffix).html(s);};")
    JS.eval("window.setJsDate = function() { canSendCommandBak = canSendCommand; canSendCommand = true; ASLEvent ('SetJsDate',  Date().toString()); canSendCommand = canSendCommandBak;};")
    JS.setJsDate()
  </function>
  <function name="SetUpStartTimeAndDate">
    timedict = GetCurrentTime()
    game.starttime = DictionaryItem (timedict, "time")
    game.startdate = DictionaryItem (GetCurrentDate(), "date")
  </function>
</asl>
```


</details>

alexwarren: It's not about the time zone, but about the time format.

For example on a machine in the US the short format for today's date is `12/27/2024`, but for the UK it's `27/12/2024`.

This is why I'm hesitant to build this in as native Quest functions because there's a _lot_ to cover. And all of this stuff exists in JavaScript already so why not just use it there.

KVonGit: Oh, I see where you're coming from. (I also learned something. I thought your format was `2024/12/27`, and I've been writing the date that way in attempt to make it easier for everyone across the pond for quite a while, haha.)

I threw the `ASLEvent` method into that example game, too. There are basically two (and a half) things to cover when using that method:

If the script is running in response to a command, `ASLEvent` will not fire online because `canSendCommand` will be `false`. I have a dirty (probably mostly unnecessary) hack in the example where I back that value up, set the variable to `true`, then change it back after calling `ASLEvent`, but I don't know how dangerous that is -- haven't had problems with it online as of yet.

The delay between JS and Quest means we can't use things like this:
```c#
game.questplatform = "notset"
JS.whereAmI()
msg(game.questplatform)
```

That will result in "notset" every time, because it depends on `ASLEvent` to set `game.questplatform`:

```js
function whereAmI(){
 ASLEvent("WhereAmI", "' + platform + '");
}
```

```xml
<function name="WhereAmI" parameters="platform">
  game.questplatform = platform
</function>
```

---
The best way to handle that delay is to use a change script, but that will fire *after* the script calling `JS.whereAmI()`, and most Quest authors don't quite grasp the concept of non-blocking scripts, let alone all this. (Not being mean here. I just recently learned what little I know about this stuff, myself.)

Also, to exacerbate things, the delay has ranged from 50ms to 2000ms when a few of us tested this a few months ago. 

---
This also has me thinking about my JS function to get the time and date for the log entries when writing the log to a file! I won't derail this issue with that, though.

Pertex: Do we have a sleep function in Quest so we could use it in a loop waiting for the ASLEvent?

KVonGit:
> 4 separate PRs please

1. Pertex - Logo and Desktop Icon - https://github.com/textadventures/quest/pull/1309
2. Pertex - endWaitDiv - https://github.com/textadventures/quest/pull/1310
3. It looks like Pertex is currently working on the Date and Time functions
4. As for the file watcher? This goes for me as well: `My knowledge is not sufficient for this`

alexwarren:
Thanks. For number 4 I think we just leave this for now.

PR #1243, opened 2024-11-27
Pertex - Add delay option to walkthrough

TextMisadventures:
Submitted by Pertex

New walkthrough step option:
`delay:`

Sets interval between steps in seconds

Example:
```
delay:3
```

After this step, each step will pause 3 seconds before calling the next step

### DIFF
--- a/Player/WalkthroughRunner.vb
+++ b/Player/WalkthroughRunner.vb
@@ -23,6 +23,7 @@ Friend Class WalkthroughRunner
     Public Sub Run()
         Dim delay As Integer = 0
         Dim dateStart = DateTime.Now
+        Dim showRuntime As Boolean = False
         For Each cmd As String In m_gameDebug.Walkthroughs.Walkthroughs(m_walkthrough).Steps
             If m_cancelled Then Exit For
             RaiseEvent MarkScrollPosition()
@@ -49,6 +50,8 @@ Friend Class WalkthroughRunner
                     Dim eventName As String = eventData(0)
                     Dim param As String = eventData(1)
                     m_game.SendEvent(eventName, param)
+                ElseIf cmd.StartsWith("runtime:") Then
+                    showRuntime = True
                 Else
                     m_game.SendCommand(cmd)
                 End If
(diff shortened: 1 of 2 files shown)


alexwarren:
What is the need for this? And why does it always need to output the duration at the end?

KVonGit:
I asked the same questions, and I was immediately outnumbered.

@Pertex will correct me if I'm wrong here, I'm sure. (The conversation must've been on my server, because I can't find it for reference now.)

Basically, I use the walkthrough for testing after making changes (and sometimes to include the steps for beta-testers), but some people like to run walkthroughs to read the text as it goes by to check for errors -- or to just watch their playthroughs.

---
(I also see a note I left myself to ask Pertex if we should/could make the runtime message optional.)

Pertex:
The point of slowing down the output is simply that I want to read what's going on. At the moment you can't see anything when the walkthrough is running. Of course you can just let it run and then scroll through, but the delay saves you all the clicking. This might not be a problem for small games, but the walkthrough of my game takes one minutes (without delay). Hence the output of the runtime at the end. Then I can immediately see whether a change to the core or game has an effect. The question is, does the output bother anyone? 

The slowdown of the walkthrough was also supported in Discord.  (and it was not KV! :-)

alexwarren:
Fair enough!

Perhaps add a separate command to output the duration?

Also, as part of this PR, could you document the new walkthrough command(s) in `/docs/using_walkthroughs.md`?

Pertex:
a separate command? good idea!

/docs/using_walkthroughs.md will be updated!

KVonGit:
> a separate command? good idea!

So, should we just add code to handle an `includeruntime:true` step? Or maybe just `runtime:true`?

KVonGit:
Just pushed a new commit with the changes.

https://github.com/textadventures/quest/pull/1243/commits/c0109c475ad06bf28c29d3021a5b18e78450fb41

(Did anyone get a notification when I did that? My other account didn't; so I don't think anyone did. If you did, let me know, and I won't post these updates when I do this in the future.)

Player/WalkthroughRunner.vb
```diff
@@ -39,11 +43,15 @@
...
                     ' ignore
+                ElseIf cmd.StartsWith("delay:") Then
+                    delay = Int32.Parse(cmd.Substring(6).Trim)
                 ElseIf cmd.StartsWith("event:") Then
                     Dim eventData As String() = cmd.Substring(6).Split(New Char() {";"c}, 2)
                     Dim eventName As String = eventData(0)
                     Dim param As String = eventData(1)
                     m_game.SendEvent(eventName, param)
+                ElseIf cmd.StartsWith("runtime:") Then
+                    showRuntime = True
```
alexwarren, line 54: If the expectation is that people use `runtime:` at the end of the walkthrough, then maybe this is where the code for displaying the runtime should be too? No need for a separate `showRuntime` boolean then, and it would mean you could use the `runtime:` command at multiple points in the walkthrough?

alexwarren (commented):
This is looking good! Just one small thought about the `runtime:` command - see separate comment.

alexwarren (approved):

PR #1305, opened 2024-12-27
Log - Restore old window functionality

### LINKED ISSUE #1025: Log
The log window has now been integrated via the user interface. Does this mean that the old log window of Quest has no function? Then I could remove the corresponding buttons and menu items from Quest.

KVonGit:
- Restore old log window functionality
  - This was removed in 5.8 because ThePix meant to to deprecate `request`
- Disable the in-game log by default
  - added editor control
    - box ticked by default because I didn't want to change the existing attribute name "nohtmllog", which was `true` by default in 5.8
- Add editor control and "writelogtofile" boolean attribute to `game` to enable writing log entries to a TXT file in "Documents\Quest Logs"
  - This feature was already in 5.8. (I only added the editor control and attribute in this PR.)
  - Set to `false` by default (unchanged)
- Update `WriteToLog` to add blank entries (to match the log window behaviour)
- Add English templates for new editor controls

---
Closes #1025 

---
image

---
image

---
image

---
image

---
[Log Window Restoration Test (aslx).zip](https://github.com/user-attachments/files/18260554/Log.Window.Restoration.Test.aslx.zip)

[Log Window Restoration Test-log.txt](https://github.com/user-attachments/files/18260530/Log.Window.Restoration.Test-log.txt)


Player/desktopplayer.js
```diff
@@ -32,11 +32,12 @@ function SaveTranscript(data) {
     transcriptString += data;
 }
 
-// Added by KV to write/append to GAMENAME-log.txt in Documents\Quest Logs
+// Write/append to GAMENAME-log.txt in Documents\Quest Logs
 function WriteToLog(data) {
-    if (!webPlayer && data != '' && typeof (data) == 'string') {
-        UIEvent("WriteToLog", getTimeAndDateForLog() + " " + data);
+    if (typeof (data) != 'string') {
+        data = '';
```
alexwarren, line 38: Do we really want to write an empty log in this case? Would it better to just return from the function and not log anything, or log some error message?

WorldModel/WorldModel/Core/CoreOutput.aslx
```diff
@@ -874,14 +874,16 @@
...
-    //request (Log, text)
+    request (Log, text)
     // Replacing double quotes with 2 single quotes
     text = Replace(text, "\"", "''")
     // Changing syntax from single quotes to escaped double quotes to allow single quotes in log entries
     if (not GetBoolean(game, "nohtmllog")){
       JS.eval("if(typeof(addLogEntry)===\"function\"){ addLogEntry(\""+text+"\"); };")
     }
-    JS.eval("if(!webPlayer && typeof(WriteToLog)===\"function\"){var s = \""+text+"\";WriteToLog(s);}")
+    if (GetBoolean(game, "writelogtofile") and StartsWith(GetFileURL("foo"),"quest")){
```
alexwarren, line 884: What's with `GetFileURL("foo")` here?

WorldModel/WorldModel/Core/CoreOutput.aslx
```diff
@@ -874,14 +874,16 @@
...
+      JS.eval("if(typeof(WriteToLog)===\"function\"){var s = \"" + text + "\";WriteToLog(s);}")
```
alexwarren, line 885: I know this is just updating what is here already, but I'm not such a fan of the `eval`s here - could this call a JS function instead?

WorldModel/WorldModel/Core/Languages/EditorEnglish.aslx
```diff
@@ -930,4 +930,6 @@
   <template name="EditorScriptObjectsHide">Hide</template>
   <template name="EditorScriptObjectsPrint">Print</template>
   <template name="EditorScriptObjectsIf">If</template>
+  <template name="EditorGameDisableInGameLog">Disable the in-game log (previously enabled by default)</template>
```
alexwarren, line 933: This feels confusing in the "Features" tab, a checkbox to disable something... plus a history lesson in brackets. Not saying I know what's better but this feels a bit clunky.

Player/desktopplayer.js
KVonGit: > Would it better to just return from the function and not log anything, or log some error message?

That's how it is before the changes, but I noticed the old log window prints those as blank entries. In fact, the blank entry was missing from the log file while I was testing something earlier, and I would have missed an error had I not seen the blank entry in the actual log window.

I'm fine with whatever you think concerning this, though.

WorldModel/WorldModel/Core/CoreOutput.aslx
KVonGit: Checking for the desktop player without using `ASLEvent`.

WorldModel/WorldModel/Core/CoreOutput.aslx
KVonGit: Should I add an `IsDesktop` function to make that IF statement look better, too? (Also, do you know of a better way to check that than `GetFileURL`?)

WorldModel/WorldModel/Core/Languages/EditorEnglish.aslx
KVonGit: Ha, yeah, I agree that it's clunky. I added the history lesson because the control just seems awkward. It's the only box that's ticked by default and disabling rather than enabling. I honestly don't like to add editor controls, but Andy usually liked to add them to make things easier to use. I'll remove the control, and I'll leave the attribute set to true to disable the in-game log (which still seems backwards).

---
Also, I'm trying to avoid saying I'd remove things, but I'd definitely remove the in-game log altogether if it were up to me. We only came up with the idea because they disabled the window, and that's back now.

Player/desktopplayer.js
alexwarren: Maybe set data to `typeof(data)`, then it's hopefully obvious to the game author that they messed up.

WorldModel/WorldModel/Core/CoreOutput.aslx
alexwarren: I feel there must be an easier way... I also feel I saw something had been added that already did something like that?

alexwarren: In any case I wonder if there's a way we can make this agnostic of where it's running.

KVonGit: > I also feel I saw something had been added that already did something like that?

You're probably thinking of `JS.whereAmI()`, and it uses `ASLEvent`.

> In any case I wonder if there's a way we can make this agnostic of where it's running.

The transcript is missing a JS function online in 5.8, and I'm beginning to notice I doubled up on trying to be sure that didn't happen again.

I could just put a "dummy" `WriteToLog` in **player.js** (which does nothing but exist), then remove all this online checking (and the `eval`, which was just a second backup for that) like this:
```js
if (GetBoolean (game, "writelogtofile")) {
      JS.writeToLog(s)
```

WorldModel/WorldModel/Core/CoreOutput.aslx
KVonGit: I got rid of all this by adding a dummy WriteToLog in player.js.

WorldModel/WorldModel/Core/Languages/EditorEnglish.aslx
KVonGit: I removed the "Disable the in-game log" control. If someone wanted to use the in-game log, they'd likely be skilled enough to set `game.nohtmllog` to `true`.

I left the "Write log to file" control in there. If you'd rather it not be there, just say the word.

I've left the in-game log in here, too, but I changed it from a JS string variable to an array (as you advised a few months ago).

KVonGit: I just uploaded a new commit, and I believe the changes cover all the issues mentioned (and more).

(I'm not sure if GitHub sends a notification when I do that.)

---
> [!NOTE]
> After I closed everything out, I realized this needs to adjust the CSS for the in-game log's dialog window when the player is using a mobile device. I will do this in the morning, and then this should be ready.

KVonGit:
I can't get the in-game log to display correctly across different screen sizes. Everything else concerning the in-game log except for the display settings works smoothly now on desktop, webplayer, and mobile, but it's no good at all if you can't see half of it.

I was trying to keep this in because ThePix really wanted online users to be able to access the log, but they can just use `JS.console.log` in place of `Log` in their scripts (or use Pixie's `DbgMsg` function).

So... I'm back to deprecating the in-game log functionality. I shall begin in 30 minutes, unless anyone objects beforehand (and I don't think anyone will).

This PR should be ready to review again after I make these changes.

KVonGit:
The changes have now been applied to this PR.

a6d6afd2

PlayerController/playercore.js
```diff
@@ -162,11 +162,11 @@ function markScrollPosition() {
 // Variable added by KV
 var gameName = document.title;
 
-function setGameName(text) {
+function setGameName(s) {
```
alexwarren, line 165: Any reason for changing this variable name here?

PlayerController/playercore.js
```diff
@@ -1302,157 +1302,53 @@ function showPopupFullscreen(title, text) {
...
-      },
-    },
-    show: { effect: "fadeIn", duration: 500 },
-    // The modal setting keeps the player from interacting with anything besides the dialog window.
-    //  (The log will not update while open without adding a turn script.  I prefer this.)
-    modal: true,
-  });
+  // In-game HTML log introduced in 5.8. Deprecated in 5.9
+  if (!displayedInGameLogWarning) {
+    console.warning("The in-game HTML log feature has been deprecated.")
```
alexwarren, line 1319: It's been fully _removed_, not just deprecated now, right?

alexwarren (commented):
Will test this properly later today, just leaving a couple of tiny comments for now.

PlayerController/playercore.js
KVonGit: I was thinking `text` is a reserved keyword in some version of ECSMAScript, but I can't find anything about that when I look it up now. I guess it's `location` I was thinking of. Someone had one of these in a game that worked in the old 5.8 browser but not online.

I'll change this back while changing "deprecated" to "removed" on line 1319.

PlayerController/playercore.js
KVonGit: Well, I'll be darned. I always thought those were synonymous. I learned something today.

I'll change this to "removed".

KVonGit: > I'll change this to "removed".

Done, and the commit has been pushed.

PlayerController/playercore.js
```diff
@@ -164,7 +164,7 @@ var gameName = document.title;
 
 function setGameName(text) {
     $("#gameTitle").remove();
-    document.title = text;
+    document.title = s;
```
alexwarren, line 167: Missed a bit! This should be reverted.
```suggestion
    document.title = text;
```

WebPlayer/playercore.js
```diff
@@ -164,7 +164,7 @@ var gameName = document.title;
 
 function setGameName(text) {
     $("#gameTitle").remove();
-    document.title = text;
+    document.title = s;
```
alexwarren, line 167: Ditto

PlayerController/playercore.js
KVonGit: Fixed. (Sorry about that. I usually run a game after making changes like this.)

PR #1312, opened 2024-12-28
new DateTime Functions

Pertex:
Hi Alex,
I changed the CurrentDate() and CurrentTime() functions to return the date / time in iso format. I added CurrentDate(int offset) and CurrentTime(int offset). With them the game creator can simulate timezones to create the local datetime of the player by adding or subtracting hours . How to get the offset? Perhaps asking the player to input his actual datetime?
Perhaps we can expand this functions with JS calls later.

Documentation  is still missing, I will add it, if this PR is acceptable

### DIFF
--- a/WorldModel/WorldModel/Functions/DateTimeFunctions.cs
+++ b/WorldModel/WorldModel/Functions/DateTimeFunctions.cs
@@ -1,0 +1,47 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace TextAdventures.Quest.Functions
+{
+    public static class DateTimeFunctions
+    {
+        public static string CurrentDate()
+        {
+            return CurrentDate(0);
+        }
+        public static string CurrentDate(int offset)
+        {
+            DateTime localDate = DateTime.Now;
+            double offsetD= Convert.ToDouble(offset);
+            localDate = localDate.AddHours(offsetD);
+            return localDate.ToString("yyyy-MM-dd");
+        }
+        public static string CurrentTime()
+        {
+            return CurrentTime(0);
+        }
+
+        public static string CurrentTime(int offset)
+        {
+            DateTime localTime = DateTime.Now;
+            double offsetD = Convert.ToDouble(offset);
+            localTime = localTime.AddHours(offsetD);
+            return localTime.ToString("HH:mm:ss");
+        }
+        
+        public static string CurrentUtcDate()
+        {
+            DateTime utcDate = DateTime.UtcNow; 
+            return utcDate.ToShortDateString();
+        }
+
+        public static string CurrentUtcTime()
+        {
+            DateTime utcTime = DateTime.UtcNow;
+            return utcTime.ToShortTimeString();
+        }
+     
+    }
+}
--- a/WorldModel/WorldModel/Core/Core.aslx
+++ b/WorldModel/WorldModel/Core/Core.aslx
@@ -432,4 +432,10 @@
     game.questplatform = s
   </function>
 
+  <type name="editor_object">
+    <isroom type="boolean">false</isroom>
+  </type>
+  <type name="editor_room">
+    <isroom />
+  </type>
 </library>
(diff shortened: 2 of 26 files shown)


WorldModel/WorldModel/Functions/DateTimeFunctions.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        public static string CurrentDate()
+        {
+            return CurrentDate(0);
+        }
+        public static string CurrentDate(int offset)
+        {
+            DateTime localDate = DateTime.Now;
+            double offsetD= Convert.ToDouble(offset);
+            localDate = localDate.AddHours(offsetD);
```
alexwarren, line 18: You don't need to do the conversion here, you can just pass the integer `offset` straight into `AddHours`.

alexwarren (changes requested):
I'm still not sure if we need these as core Quest functions. Perhaps a small sample to demonstrate why these are useful and how they can be used? Then we can experiment with other implementations like doing it entirely in JavaScript.

Also if the intention of the offsets is to simulate timezones, I'd expect these to be on the UTC versions, not the local timezone ones?

WorldModel/WorldModel/Functions/DateTimeFunctions.cs
Pertex: That's right. I tried to use optional parameter like int? offset=null
There it said its not possible to convert it to double

Pertex:
In my game I call the functions in the inituserinterface to set the start time when the player starts to play his game. When he finishes the game I call the function again and save this value. With both values I calculate how many days he has been playing the game.
Its very easy to use them in Quest to get time and date informations. Or, for example, to make a game that runs for exactly two hours in real time. 
We have also thought about JS, but have not found a solution that is easy to use. One call would probably be JS.getDate(), for example. Then you would have to end the current function and wait for JS to send an ASLEvent with the result to a quest function. Now the actual function must be continued again in order to evaluate the result.
I have not found a quick and simple solution for this that even a beginner can understand.
Do you have any ideas on how to proceed? Could there be a solution where Quest stops until a JS response is received?

Pertex:
UTC version? Sorry, I have ignored them. I would also install the offset variant if DateTime were included in Quest

KVonGit:
> Perhaps a small sample to demonstrate why these are useful and how they can be used? Then we can experiment with other implementations like doing it entirely in JavaScript.

If I were to use this in a game:

<details><summary> Example game </summary>

[Do You Have the Time (For the Quest 5.9 testing build only).zip](https://github.com/user-attachments/files/18263949/Do.You.Have.the.Time.For.the.Quest.5.9.testing.build.only.zip)

```xml
<asl version="580">
  <include ref="Core.aslx" />
  <game name="Do You Have the Time?">
    <gameid>eda4b815-d1ee-4203-b442-d0559f12654d</gameid>
    <version>1.0</version>
    <feature_advancedscripts />
    <attr name="autodescription_youarein_useprefix" type="boolean">false</attr>
    <attr name="autodescription_description" type="int">2</attr>
    <attr name="autodescription_youcango" type="int">4</attr>
    <attr name="autodescription_youcansee" type="int">3</attr>
    <autodescription_description_newline />
    <inituserinterface type="script">
      SetUpInitTimeAndDate
    </inituserinterface>
    <start type="script">
      SetUpStartTimeAndDate
    </start>
  </game>
  <object name="room">
    <isroom />
    <description><![CDATA[<br/>The current time is {=DictionaryItem (GetCurrentTime(),"time")}{if not game.islocaltime: UTC}, and the date is {=DictionaryItem (GetCurrentDate(), "date")}.<br/><br/>JavaScript Date(): <span id="js-date-{game.currentturnoutputsection}"><img src="{=GetFileURL("foo")}" onerror="updateJsDateDiv('{game.currentturnoutputsection}');$(this).remove();"/>loading...</span><br/><br/>You started this game at {game.starttime} on {game.startdate}.<br/><br/>You loaded the current session at {game.inittime} on {game.initdate}.<br/><br/>Wait until the time changes and type {command:LOOK} to test this room description.<br/><br/>Use the {command:SAVE} command, then load the saved game to test the start and init time and date attributes on the game object.<br/><br/>]]></description>
    <alias>Time and Date Testing Grounds</alias>
    <usedefaultprefix type="boolean">false</usedefaultprefix>
    <object name="player">
    </object>
  </object>
  <function name="IsDesktop" type="boolean">
    return (GetFileURL("BaseURI").StartsWith("quest"))
  </function>
  <function name="GetCurrentTime" type="dictionary">
    if (IsDesktop()) {
      return (QuickParams ("time", CurrentTime(),"local", true))
    }
    return (QuickParams ("time", CurrentUtcTime(),"local", false))
  </function>
  <function name="GetCurrentDate" type="dictionary">
    if (IsDesktop()) {
      return (QuickParams ("date", CurrentDate(), "local", true))
    }
    return (QuickParams ("date", CurrentUtcDate(), "local", false))
  </function>
  <function name="SetJsDate" parameters="jsdate"><![CDATA[
    game.changedjsdate => {
      JS.updateJsDateDiv (game.currentturnoutputsection, game.jsdate)
    }
    game.jsdate = jsdate
  ]]></function>
  <function name="SetUpInitTimeAndDate">
    timedict = GetCurrentTime()
    game.inittime = DictionaryItem (timedict, "time")
    game.islocaltime = DictionaryItem (timedict, "local")
    game.initdate = DictionaryItem (GetCurrentDate(), "date")
    JS.eval("window.updateJsDateDiv = function(idsuffix, s=Date().toString()) {$('#js-date-' + idsuffix).html(s);};")
    JS.eval("window.setJsDate = function() { canSendCommandBak = canSendCommand; canSendCommand = true; ASLEvent ('SetJsDate',  Date().toString()); canSendCommand = canSendCommandBak;};")
    JS.setJsDate()
  </function>
  <function name="SetUpStartTimeAndDate">
    timedict = GetCurrentTime()
    game.starttime = DictionaryItem (timedict, "time")
    game.startdate = DictionaryItem (GetCurrentDate(), "date")
  </function>
</asl>
```

</details>

---
I threw the `ASLEvent` method into that example game, too. (It's the best approach anyone has come up with that I've seen. (Using a change script.))

alexwarren:
It sounds from Pertex's comment like the use case is the ability to time how long a game session has taken, in real time (so the clock keeps ticking even when the game is not active). That way the length can be reported (like in an end game message), or the game can end after 2hrs etc.

In that case I'm not sure the back-end (i.e. the game's ASL code) needs to worry about time zones and UTC or formatting? You'd have to do some string manipulation and maths with the current PR to extract the data you need.

Perhaps the simplest way to support the use case would be one function which returns the current Unix epoch time in UTC? Then it's just an integer and you can calculate durations from that.

I think anything that needs a formatted or local time/date could be done entirely in JS, but let me know if there are other use cases I'm missing.

Pertex:
>Perhaps the simplest way to support the use case would be one function which returns the current Unix epoch time in UTC?

This would be a Quest function or a JS function?

alexwarren:
A Quest function.

KVonGit:
I think Andy already wrote a Quest function like that. It"s either somewhere in his GutHub wiki, or in an old forum post. I'll see if I can find it.

KVonGit:
https://github.com/ThePix/quest/blob/master/ClockLib.aslx

This is not exactly what Alex is talking about, but some of the code in it might help you a little.

PR #1313, opened 2024-12-28
Add note about using HTML in ShowMenu options [DOCS]

KVonGit:
https://discord.com/channels/1278365322054996018/1312531319791681617/1322678069344862252

The WebPlayer app can't handle HTML (`<`) in an option's string when selected. However, a dictionary of options can be used as a workaround.


alexwarren:
I feel it would be better to fix the underlying bug here? Using HTML _should_ work, I haven't tested it but my suspicion would be one of ASP.NET's "helpful" features to prevent script injection attacks - in which case there will be some attribute we can use to disable this.

KVonGit:
Sound great (and like it might solve the similar issue with the WebEditor too).

Closing this and creating an Issue instead.

PR #1316, opened 2024-12-29
Pertex Fixes AllRooms

KVonGit:
Sets `isroom` to `true` when creating a room, or when using the dropdown to change to "room" or "Object and room".

Changed "Object and/or room" to "Object and room"

Pertex contributed the functional part of this code, and I merely changed the editor controls and such


EditorControls/DropDownTypesControl.xaml.cs
```diff
@@ -116,6 +116,14 @@ public void Save()
...
             m_helper.Controller.StartTransaction(String.Format("Change type from '{0}' to '{1}'", m_dropDownValues[m_currentType], m_dropDownValues[selectedType]));
 
+            if (selectedType == k_noType)
+            {
+              if (m_currentType == "editor_object" ||  m_currentType == "editor_room")
+                {
+                  m_helper.Controller.AddFieldToElement(m_data.Name, "isroom", true);
+                }
+            }
+
```
alexwarren, line 126: This is feeling clunky to me. This is changing the dropdown types control to have some special cases to set other fields on an element. It's a sign that something is wrong with this design because logically the drop down types control shouldn't need to know anything about what a room is, or changing other fields on an element.

WorldModel/WorldModel/Core/CoreEditor.aslx
```diff
@@ -15,11 +15,9 @@
...
-
-  <type name="editor_room">
-    <isroom />
-  </type>
-  <type name="editor_object"/>
+  
+  <type name="editor_room"/>
```
alexwarren, line 19: The core of the issue we're trying to fix is that these editor types get stripped out when creating a `.quest` file right?

It seems to be that the better simpler fix would be to have "actual" types to handle room vs non-room?

alexwarren (changes requested):

WorldModel/WorldModel/Core/CoreEditor.aslx
Pertex: May you plz explain "actual types"?

alexwarren: I mean non-editor types.

KVonGit: > May you plz explain "actual types"?

I think he means we should create a `defaultroom` type with `is_room` set to `true`. It would only be a default object, and `is_room` would be its only additional attribute.

So, in this bit of code we should also add a new `defaultroom` type, and [the other bit of code in DropDownTypesControl](https://github.com/textadventures/quest/pull/1316#discussion_r1899175100) would add the `defaultroom` type rather than setting `is_room`?

alexwarren: There shouldn't need to be any changes to DropDownTypesControl if this is done right, I think - it should be possible to handle this entirely with inherited types.

Pertex: How would this work? an object is from type defaultobject, a room could be from a new type defaultroom and a new type defaultobjectroom for "is room and object"?

alexwarren: OK I refamiliarised myself with the code.

`editor_room` and `editor_object` don't get included in `.quest` files because they're defined in `CoreEditor.aslx`, which has `type="editor"` which causes it not to be included in packaged `.quest` files.

So I think the absolute simplest fix here is to move those types into `Core.aslx`. That way there's no need to change or update anything else.

KVonGit: Gotcha. Closing this and creating a new PR without all these commits in the history.

Pertex: @KVonGit Do you open a new issue for this plz?
And keep the changed code! Could there be problems with existing games? Which type is for "is object and room"?

KVonGit:
> And keep the changed code! 

I gotcha.

> Could there be problems with existing games?

Negative. 😁   👇 

> Which type is for "is object and room"?

`editor_room`

It will carry over into published games after the change.

KVonGit:
> `object and room"?

Oh. 🤔 

I *think*:
- object
  - editor_object
 - room
   - editor_room
 - object and room
   - editor_object
   - editor_room

Pertex:
object and room

    editor_object
    editor_room

as far as I know only one type can be inherited

KVonGit:
The `editor_*` types only tell the editor what controls to display.

The only exception is what was added in 5.8 to `editor_room`, which is `is_room`.

Once we add `editor_room` to **Core.aslx**, that will carry over to published games.

But would this mean the drop-down control should do anything differently than it does in 5.8?

And what about the wonky tick-box in 5.8?

image

I bet a fake nickel that moving `editor_room` to **Core.aslx** makes it work just like it does after your changes, but without messing with the drop-down __or__ needing that check-box. I'm testing it in 5, 4, 3, 2...

KVonGit:
> object and room
> 
> ```
> editor_object
> editor_room
> ```
> 
> as far as I know only one type can be inherited

`:(`

Pertex:
> And what about the wonky tick-box in 5.8?
> 
> image
> 

Must be removed because this checkbox is *'+#§&

KVonGit:
Well... It sets `is_room` to `true` when you used the drop-down to change an object to a room. It also works as expected when changing from a room to an object with the drop-down.

...but, as Pertex already knew, it does not work when switching to or from "object and room".

That is because it does not add the `editor_room` type in this case. In fact, this removes `editor_object` ***and*** `editor_room`.

The editor controls in that use these types:

```xml
<!-- EditorObjectRoom.aslx -->
<mustnotinherit>editor_object</mustnotinherit>
```

```xml
<!-- EditorObjectObject.aslx -->
<mustnotinherit>editor_room</mustnotinherit>
```

KVonGit:
That's why you had to go back and make an extra change to the drop-down control. I'd forgotten about that.

KVonGit:
So, we want the drop-down control to add the `editor_room` type when it's "object and a room", then everything should be fine (I think).

PR #1320, opened 2024-12-29
Add MovePlayer function

### LINKED ISSUE #1297: Add "move player" function
Currently to move the player you have to move game.pov, and this is commonly confused with the ChangePOV function. So let's add a "move player" function to live alongside "move object", which should be nice and easy to find.

KVonGit:
Closes #1297

> Currently to move the player you have to move game.pov, and this is commonly confused with the ChangePOV function. So let's add a "move player" function to live alongside "move object", which should be nice and easy to find.
- Alex

### DIFF
--- a/WorldModel/WorldModel/Core/CoreEditorScriptsObjects.aslx
+++ b/WorldModel/WorldModel/Core/CoreEditorScriptsObjects.aslx
@@ -21,6 +21,24 @@
   </editor>
   
   <editor>
+    <appliesto>(function)MovePlayer</appliesto>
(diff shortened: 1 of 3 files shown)


KVonGit:
This modifies **EditorEnglish.aslx**, and so does #1316 

Closing this, to resubmit after the previous PR is finalized.

KVonGit:
#1316 has been closed without merging.

So, reopening this.

alexwarren:
I don't think there would have been a conflict here by the way, as they were two different parts of the same file, and git is usually pretty good at handling that situation.

alexwarren (approved):

PR #1329, opened 2024-12-31
Save the outputData element when clearing the screen

KVonGit:
- Remove "scrollback" HTML elements created for dialog when closing dialog to avoid errors, as they each have an ID
- Change "scrollback" width and height settings to be slightly smaller than the inner-dimensions of the window
- Move `#outputData` out of `#divOutput` before clearing the screen, then back again at the end of `clearScreen`. (To retain the data when saving/resuming online)

Related: #1323 

Note: I can split these up if that would be better.

Also note: this does not close issue 1200, and I can't edit that commit message. 1200 needs code modified in the save command's script -- or `sendCommand` in the WebPlayer needs to catch the save command and call `saveGame` from there (or both?)

### DIFF
--- a/PlayerController/playercore.js
+++ b/PlayerController/playercore.js
@@ -768,6 +768,8 @@ function disableAllCommandLinks() {
 var saveClearedText = false;
 var clearedOnce = false;
 function clearScreen() {
+    // Move outputData out of divOutput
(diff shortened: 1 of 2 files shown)


KVonGit:
I'm redoing this as two PRs, and without closing the wrong issue!

PR #1331, opened 2024-12-31
Retain outputData during call to clearScreen

### LINKED ISSUE #1314: JS getCurrentDiv function
When loading a saved game, `getCurrentDiv` does not always work properly.

https://discord.com/channels/1278365322054996018/1278365322763567199/1311400615825641532

I will submit a PR for this. It is in the PR queue, haha.

KVonGit:
- Move the `#outputData` element out of the `#divOutput` element before dumping the data with `clearScreen`
- Set ID of all elements with the `.clearedScreen` class to `null`
  - Otherwise lines appear at the top of the page due to duplicate or *de facto* sequential IDs, and other shenanigans
- Return `#outputData` to `#divOutput` before exiting `clearScreen` script

Closes #1314

### DIFF
--- a/PlayerController/playercore.js
+++ b/PlayerController/playercore.js
@@ -768,6 +768,8 @@ function disableAllCommandLinks() {
 var saveClearedText = false;
 var clearedOnce = false;
 function clearScreen() {
+    // Move outputData out of divOutput
+    $("#outputData").appendTo($("body"));
     if (!saveClearedText) {
         $("#divOutput").css("min-height", 0);
         $("#divOutput").html("");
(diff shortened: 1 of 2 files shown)


KVonGit:
I fixed the error.

(I also just changed my browser settings so there is no cache. I had this change in the testing build last week. Sometimes, JS files don't refresh as expected when I'm running the WebPlayer app after making changes. This one was one of those cases. My cache had a **playercore.js** from 3 builds ago when I checked in the debugger!)

KVonGit:
Closing this and submitting the transcript PR. I'll redo this one after that one.

PR #1344, opened 2025-01-01
Allow alt names for exits

### LINKED ISSUE #529: Allow alt names for exits
I can do this myself, but I had to find out how to do it, and it didn't work
the same as for objects. I had a case where I had an exit called "men's room",
and I wanted to allow the player to type "mens room" as well (since some
would). I won't go into what I tried, but in the end, I was able to create an
"alt" attribute on the exit and provide the alternate names, and it worked.  

The difficulties:  
1) It's not parallel to creating "alt" for an object, since "alt" for an
object is "implied" to be a list, but it didn't seem to be for an exit. (Does
this mean an exit is not an object?) I had to explicitly put the "list" type
on the alt tag. It was a small detail, but it gave me a strange error message
until I did since "foreach" on a string goes character by character, and it
was expecting a list of strings...  

2) Well, it didn't show up in the UI. :)  

This isn't high priority, but it might be handy for those times when a game
creator does want multiple names for an exit.  

Todo:  
1) Add an "implied" for "alt" as a list on an exit. (I haven't tried this.)  
2) Add the "alt" field into the UI.  
## Work Item Details

**Original CodePlex Issue:**    [Issue 1155](http://quest.codeplex.com/workitem/1155)
**Status:** Proposed
**Reason Closed:**  Unassigned
**Assigned to:**    Unassigned
**Reported on:**    Sep 28, 2012 at 11:46 AM
**Reported by:**    jaynabonne
**Updated on:** May 2, 2013 at 9:04 AM
**Updated by:** alexwarren

KVonGit:
Closes #529

<details><summary> Screenshots </summary>

### Default north
image

---
### Add 'outside' alt
image

---
### Default south
image

---
### Add 'inside' alt
image

</details>

---
<details><summary> <b>Alts for Exits.aslx</b> </summary>

```xml
<asl version="580">
  <include ref="Core.aslx" />
  <game name="Alts for Exits">
    <gameid>4f0de892-13f4-4216-820c-6d679f5b88f5</gameid>
    <version>1.0</version>
    <firstpublished>2025</firstpublished>
  </game>
  <object name="room">
    <isroom />
    <object name="player">
    </object>
    <exit alias="north" to="outside">
      <inherit name="northdirection" />
      <alt type="stringlist">
        <value>n</value>
        <value>outside</value>
      </alt>
    </exit>
  </object>
  <object name="outside">
    <isroom />
    <exit alias="south" to="room">
      <inherit name="southdirection" />
      <alt type="stringlist">
        <value>s</value>
        <value>inside</value>
      </alt>
    </exit>
  </object>
</asl>
```

</details>

---
<details><summary> Transcript </summary>

```
Alts for Exits

You are in a room.
You can go north.

> outside
I don't understand your command.

> go outside

You are in an outside.
You can go south.

> inside
I don't understand your command.

> in

You are in a room.
You can go north.

> out

You are in an outside.
You can go south

> go south

You are in a room.
You can go north.

> n

You are in an outside.
You can go south.

> s

You are in a room.
You can go north.

> go n

You are in an outside.
You can go south.

> go s

You are in a room.
You can go north.

> go outside

You are in an outside.
You can go south.

> go inside

You are in a room.
You can go north.
```

</details>

### DIFF
--- a/WorldModel/WorldModel/Core/Core.aslx
+++ b/WorldModel/WorldModel/Core/Core.aslx
@@ -12,6 +12,7 @@
   <implied element="timer" property="interval" type="int"/>
   <implied element="timer" property="script" type="script"/>
   <implied element="turnscript" property="script" type="script"/>
+  <implied element="exit" property="alt" type="simplestringlist"/>
 
   <include ref="CoreCommands.aslx"/>
   <include ref="CoreFunctions.aslx"/>
--- a/WorldModel/WorldModel/Core/CoreEditorExit.aslx
+++ b/WorldModel/WorldModel/Core/CoreEditorExit.aslx
@@ -21,6 +21,13 @@
       </control>
 
       <control>
+        <caption>[EditorObjectObjectOthernames]</caption>
+        <controltype>list</controltype>
+        <attribute>alt</attribute>
+        <editprompt>[EditorObjectObjectPleaseenter]</editprompt>
+      </control>
+
+      <control>
         <caption>[EditorExitAlias]</caption>
         <controltype>dropdown</controltype>
         <attribute>alias</attribute>


KVonGit:
I added an 'outside' alt to a north exit.

The command `OUTSIDE` does not work, but `OUT` does work. In fact, everything I threw at it except for `OUTSIDE` worked.

- N
- NORTH
- GO N
- GO NORTH
- GO TO NORTH
- OUT
- GO OUT
- GO TO OUT
- GO OUTSIDE
- GO TO OUTSIDE

---
But if I add a second exit from the main room, and I make it an "out" exit, there is no disambiguation menu when I enter `OUT`. It just defaults to the actual "out" exit.

ROOM
- n
  - to back yard
  - alt: outside
- out
  - to front yard
  - no alt added

```
Alts for Exits

You are in a room.
You can go north or out.

> outside
I don't understand your command.

> out

You are in a front yard.
You can go in.

> in

You are in a room.
You can go north or out.

> go outside

You are in a back yard.
You can go south.
```

---
This technically works, but it doesn't quite work as expected. Back to the drawing board!

angelwedge:
The pattern for the `go` command specifically matches the literal `north|east|south|west|northeast|northwest|southeast|southwest|in|out|up|down|n|e|s|w|ne|nw|se|sw|o|u|d` as single-word commands; so those are the only ones that work without 'go'. If you want it to work for "outside", you would need to add that to the pattern.

> But if I add a second exit from the main room, and I make it an "out" exit, there is no disambiguation menu when I enter OUT

There wouldn't be. The game only considers partial matches if there isn't an exact match. Although it might be surprising in some contexts, this is the intended behaviour.

KVonGit:
> If you want it to work for "outside", you would need to add that to the pattern.

Ah... I see!

So, that means it's no good unless the `go` command is modified to check the `alt` attributes of all the visible exits in scope to allow this behaviour. (Or am I the only one who would expect the `alt` to work on its own, without adding "go" or "go to"?)

---
> The game only considers partial matches if there isn't an exact match.

I've been known to turn that off. In fact, I think you gave me code to make it match at least six characters[^1], like old DOS games.

[^1]: Unless the name had less than six characters, of course. `:p`

angelwedge:
> So, that means it's no good unless the go command is modified to check the alt attributes of all the visible exits in scope to allow this behaviour. (Or am I the only one who would expect the alt to work on its own, without adding "go" or "go to"?)

The go command can't check the alt attribute in this case; because the parser only knows that it's the "go" command if the command is one of those specific words, and trusts the command pattern to tell it whether the parameter is an object, exit, or text.

I've seen people comment on it on the forums, and the answer is that typing the exit name without "go" only works for those specific names; and if you have other exit names you'll have to manually exit them into the regex. This isn't even something that can be easily automated, because the regex for each command is compiled the first time the parser runs, and subsequent changes to a command's pattern have no effect.

You could give the go command the pattern `#exit#` - but then it would fire for every unknown command, and so it would tell the player "I can't go that way" instead of "unrecognised command".

You could use `game.unresolvedcommandhandler` to check whether the command is the name of an exit in scope, and to call HandleSingleCommandPattern`; but that breaks in the case where the user has entered "outside. get fish. north" (if 'fish' needs a disambiguation menu, it would execute the third command while waiting on disambiguation for the second).

I remember more than once saying on the forums that there's no sane way to make this work… but now I'm not so sure. Off the top of my head, this `game.unresolvedcommandhandler` might help:
```
// check if the command entered is a bare exit name
// (we can't use the core ResolveNextName because we don't want it to give an error message if there's no match)
command = LCase(command)
fullmatches = NewObjectList()
partialmatches = NewObjectList()

foreach (exit, ScopeExits()) {
  name = LCase(GetDisplayAlias(exit))
  CompareNames (name, value, exit, fullmatches, partialmatches)
  if (EndsWith (TypeOf (exit, "alt"), "list")) {
    foreach (altname, exit.alt) {
      CompareNames (LCase(altname), value, exit, fullmatches, partialmatches)
    }
  }
}
match = null
if (ListCount (fullmatches) = 1) {
  match = ListItem (fullmatches, 0)
}
else if (ListCount (fullmatches) = 0 and ListCount (partialmatches) = 1) {
  match = ListItem (partialmatches, 0)
}
if (not match = null) {
  // The command matches the alias or alt of exactly one exit!
  // back up the command queue, so that our call to HandleSingleCommandPattern won't try to recurse onto the next item
  queue = game.pov.commandqueue
  game.pov.commandqueue = null

  // We let the command resolve its exit normally
  // we know it won't trigger disambiguation because we don't allow control to pass here unless there's a single match
  HandleSingleCommandPattern (go, QuickParams ("exit", command))

  // And now, restore the command queue so that the subsequent call to `HandleNextCommandQueueItem` will work as normal
  game.pov.commandqueue = queue
}
else {
  msg (Template("UnrecognisedCommand"))
}
```

Something similar could probably be added to `HandleSingleCommand`, if we decide that's an intuitive way for the engine to work. But in that case, you'd want to add it as an extra `if (thiscommand = null) {` before the existing one. In that case, it might actually be beneficial to have it consider visible objects too. If the command entered is just the name of an exit, trigger "go". If it's an object, look for a `defaultcommand` attribute, and pass it to the referenced command (so that if there is an object for which this behaviour seems intuitive, the game designer can choose the command). If an object without the attribute, display "What do you want to do with it?". And if it doesn't match, allow the existing unresolved command behaviour to continue.

PR #1347, opened 2025-01-02
Pertex - Add custom location bar background image option

KVonGit:
Submitted by Pertex

Add option to use custom image in the location bar, complete with editor controls

Can use URL or local image

<details><summary> Screenshot </summary>

image

</details>

<details><summary> <b>PertexTester.aslx</b> </summary>

```xml
<asl version="580">
  <include ref="Core.aslx" />
  <game name="pertex tester">
    <gameid>af60d15c-82de-4042-b14a-432ba7ac84e6</gameid>
    <version>1.0</version>
    <classiclocation type="boolean">false</classiclocation>
    <locationbarimage>https://secure.gravatar.com/avatar/4cc1e52f99452e841ecf4cbee9eadde9?d=retro</locationbarimage>
    <locationbarimageisurl />
    <showlocation />
    <customlocationtextcolour>Transparent</customlocationtextcolour>
  </game>
  <object name="room">
    <isroom />
    <object name="player">
    </object>
  </object>
</asl>
```

</details>

### DIFF
--- a/WorldModel/WorldModel/Core/Core.aslx
+++ b/WorldModel/WorldModel/Core/Core.aslx
@@ -126,14 +126,13 @@
       else {
         JS.setCss ("#status", "border:none")
       }
-      // Pertex: this is new
       if (HasString(game, "locationbarimage")){
         url = game.locationbarimage
         if (not Trim(url) = ""){ 
-          if (not StartsWith(game.locationbarimage,  "http")){
-            url = GetFileURL(game.locationbarimage)
+          if (not StartsWith(url,  "http://") and not StartsWith(url, "https://")){
+            url = GetFileURL(url)
+            JS.eval ("$('.ui-widget-header').css('background-image','url(" + url + ")');")
           }
-          JS.eval ("$('.ui-widget-header').css('background-image','url(" + url + ")');")
         }
       }
     }
--- a/WorldModel/WorldModel/Core/Core.aslx
+++ b/WorldModel/WorldModel/Core/Core.aslx
@@ -127,12 +127,9 @@
         JS.setCss ("#status", "border:none")
       }
       if (HasString(game, "locationbarimage")){
-        url = game.locationbarimage
-        if (not Trim(url) = ""){ 
-          if (not StartsWith(url,  "http://") and not StartsWith(url, "https://")){
-            url = GetFileURL(url)
-            JS.eval ("$('.ui-widget-header').css('background-image','url(" + url + ")');")
-          }
+        if (not Trim(game.locationbarimage) = ""){ 
+          url = GetFileURL(game.locationbarimage)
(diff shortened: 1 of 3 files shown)


WorldModel/WorldModel/Core/Core.aslx
```diff
@@ -126,6 +126,16 @@
       else {
         JS.setCss ("#status", "border:none")
       }
+      // Pertex: this is new
```
alexwarren, line 129: You don't need to include comments like this in-line, we can always use Git to see the history of a file.

WorldModel/WorldModel/Core/Core.aslx
```diff
@@ -126,6 +126,16 @@
       else {
         JS.setCss ("#status", "border:none")
       }
+      // Pertex: this is new
+      if (HasString(game, "locationbarimage")){
+        url = game.locationbarimage
+        if (not Trim(url) = ""){ 
+          if (not StartsWith(game.locationbarimage,  "http")){
```
alexwarren, line 133: This should check if the url starts with `http://` or `https://`

alexwarren (changes requested):
Left a couple of comments.

I'm also wondering if we actually really want the ability to load from an external URL? I think it's better for games to include all the resources they need within the game itself?

WorldModel/WorldModel/Core/Core.aslx
KVonGit: We removed the ability to use an external URL.

If `game.locationbarimage` is an empty string (or starts with `http://` or `https://`), nothing will happen in this bit of the script now.

WorldModel/WorldModel/Core/Core.aslx
```diff
@@ -126,6 +126,15 @@
...
         JS.setCss ("#status", "border:none")
       }
+      if (HasString(game, "locationbarimage")){
+        url = game.locationbarimage
+        if (not Trim(url) = ""){ 
+          if (not StartsWith(url,  "http://") and not StartsWith(url, "https://")){
+            url = GetFileURL(url)
+            JS.eval ("$('.ui-widget-header').css('background-image','url(" + url + ")');")
+          }
+        }
```
alexwarren, line 136: `url` isn't an accurate variable name here outside the `if`. Also, I guess we don't actually need the `http://` check? Not sure what happens if you pass a URL into `GetFileURL`... either way I presume we don't do a check like this anywhere else the author can supply images.

```suggestion
        if (not Trim(game.locationbarimage) = ""){ 
          url = GetFileURL(game.locationbarimage)
          JS.eval ("$('.ui-widget-header').css('background-image','url(" + url + ")');")
        }
```

KVonGit:
I applied your suggested changes directly from your comment. (Not sure if that sends a notification.)

alexwarren (approved):

PR #1348, opened 2025-01-02
function CurrentDateUTC returns the current Unix epoch time in UTC

Pertex:
function CurrentDateUTC() returns the current Unix epoch time in UTC
documentation will follow, when this is ok

### DIFF
--- a/WorldModel/WorldModel/Functions/DateTimeFunctions.cs
+++ b/WorldModel/WorldModel/Functions/DateTimeFunctions.cs
@@ -7,9 +7,9 @@ namespace TextAdventures.Quest.Functions
 {
     public static class DateTimeFunctions
     {
-        public static long CurrentDateUTC()
+        public static string CurrentDateUTC()
         {
-            return DateTimeOffset.UtcNow.ToUnixTimeSeconds();
+            return DateTimeOffset.UtcNow.ToUnixTimeSeconds().ToString();
         }
 
     }
--- a/WorldModel/WorldModel/Functions/DateTimeFunctions.cs
+++ b/WorldModel/WorldModel/Functions/DateTimeFunctions.cs
@@ -1,0 +1,16 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace TextAdventures.Quest.Functions
+{
+    public static class DateTimeFunctions
+    {
+        public static long CurrentDateUTC()
+        {
+            return DateTimeOffset.UtcNow.ToUnixTimeSeconds();
+        }
+
+    }
+}
--- a/docs/functions/currentDateUTC.md
+++ b/docs/functions/currentDateUTC.md
@@ -1,0 +1,12 @@
+---
+layout: index
+title: CurrentDateUTC
+---
+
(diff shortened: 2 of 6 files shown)


alexwarren (approved):
Looks good!

WorldModel/WorldModel/Functions/DateTimeFunctions.cs
```diff
@@ -7,9 +7,9 @@ namespace TextAdventures.Quest.Functions
 {
     public static class DateTimeFunctions
     {
-        public static long CurrentDateUTC()
+        public static string CurrentDateUTC()
```
alexwarren, line 10: Is this because Quest doesn't support `long`s? If so doesn't that make doing calculations based on this difficult?

Pertex:
I changed the return value to string. IsInt and ToInt also work with it. From 2037, we will have to consider whether to introduce the long data type for this, because then the timestamp integer will overflow

WorldModel/WorldModel/Functions/DateTimeFunctions.cs
Pertex: you are too fast :-)

Pertex: Are long calculations so rare that they don't appear in Quest yet? I do not assume that there will be large calculations for the future with CurrentDateUTC(). What would have to be done to fully supplement long? Otherwise you could, for example, subtract 1000 years from the timestamp and use it for intermediate calculations.

I'll also have a look at the JS side to see if there are any problems

WorldModel/WorldModel/Functions/DateTimeFunctions.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Text;
+
+namespace TextAdventures.Quest.Functions
+{
+    public static class DateTimeFunctions
+    {
+        public static string CurrentDateUTC()
+        {
+            return DateTimeOffset.UtcNow.ToUnixTimeSeconds().ToString();
+        }
```
alexwarren, line 13:
```suggestion
        public static int CurrentDateUTC()
        {
            return (int)(DateTimeOffset.UtcNow.ToUnixTimeSeconds());
        }
```

alexwarren:
Let's cast this to an `int` for now, as the game author would have to do that anyway if it's a string here. We can make Quest support `long` by 2037 (maybe we'll make all integers long integers)

docs/functions/currentDateUTC.md
```diff
@@ -0,0 +1,12 @@
+---
+layout: index
+title: CurrentDateUTC
+---
+
+    CurrentDateUTC()
+
+**New in Quest 5.9** 
+Returns the [string](../types/string.html) value of the current Unix epoch time in UTC. 
+When the game is started in the desktop player, the timestamp of the player's time zone is displayed. If the game is played via the browser over the Internet, the timestamp of the web server is output.
```
alexwarren, line 10: The player's timezone shouldn't matter if this is returning UTC?

alexwarren (approved):

Issue #1200, opened 2024-10-04
WebPlayer - SAVE Command does not save divOutput


KVonGit:
Online, if you click the Save button, everything works correctly when you load your saved game.

However, if you enter the SAVE command instead, the game's text is not restored when you load your saved game.

Looks like a possible fix was posted in this thread: https://textadventures.co.uk/forum/quest/topic/5fd2-liccuidcbbf_eiefa/log-im-trying-to-figure-out-how-to-save-the-content-to-a-file#f91f54ac-f3e9-480a-b98b-83db552b4b4e

If I find a fix during testing, I will create a pull request.

KVonGit:
PR coming as soon as **playercore.js** is "free"

```js
// Modified by KV to handle the scrollback feature
var saveClearedText = false;
var clearedOnce = false;
function clearScreen() {
    // Move outputData out of divOutput
    $("#outputData").appendTo($("body"));
    if (!saveClearedText) {
        $("#divOutput").css("min-height", 0);
        $("#divOutput").html("");
        createNewDiv("left");
        beginningOfCurrentTurnScrollPosition = 0;
        setTimeout(function () {
            $("html,body").scrollTop(0);
        }, 100);
    } else {
        $("#divOutput").append("<hr class='clearedAbove' />");
        if (!clearedOnce) {
            addText('<style>#divOutput > .clearedScreen { display: none; }</style>');
        }
        clearedOnce = true;
        $('#divOutput').children().addClass('clearedScreen');
        $('.clearedScreen').attr('id',null);
        $('#divOutput').css('min-height', 0);
        createNewDiv('left');
        beginningOfCurrentTurnScrollPosition = 0;
        setTimeout(function () {
            $('html,body').scrollTop(0);
        }, 100);
    // Move outputData back to divOutput
    $("#outputData").appendTo($("#divOutput"));
    }
}
```

KVonGit:
I can't trace what's happening in between entering the SAVE command in the web player and `saveGameResponse()` being called.

- enter SAVE and press Enter
- `runCommand()` \<-- playercore.js
- `sendCommand('save')` \<-- playercore.js
- `request("RequestSave", "")` \<-- CoreCommands.aslx
  - this seems to be where we lose the data from divOutput?
- `m_worldModel.PlayerUI.RequestSave(null);` \<-- RequestSaveScript.cs
- `m_game.Save(m_saveFile, html)` \<-- Player.cs
[...]

PR #1372, opened 2025-01-15
Save command will use JS.saveGame online

KVonGit:
When saving progress online, the text from `#divOutput` is not retained when calling the `request` script.

In 5.8, we changed it to use the game object attribute set by `JS.whereAmi()`, which ended up being unused (due to issues when run from the init script). This script initially included a call to whereAmI if the attribute was not set, which was also problematic, and it devolved into its current state -- which does not work on its own.

This will not use JS to check the platform and will still use JS to save when online.

### DIFF
--- a/WorldModel/WorldModel/Core/CoreCommands.aslx
+++ b/WorldModel/WorldModel/Core/CoreCommands.aslx
@@ -923,7 +923,7 @@
   <command name="save">
     <pattern type="string">^save$</pattern>
     <script>
-      if (IsDesktop()){}
+      if (IsDesktop()){
         request(RequestSave, "")
       }
       else {
--- a/WorldModel/WorldModel/Core/CoreCommands.aslx
+++ b/WorldModel/WorldModel/Core/CoreCommands.aslx
@@ -913,7 +913,6 @@
   <command name="wait" pattern="[wait]">msg (Template("DefaultWait"))</command>
   <command name="xyzzy" pattern="[xyzzy]">msg (Template("DefaultXyzzy"))</command>
   
-  <!-- Modified to suppress turn scripts -->
   <command name="help" pattern="[help]">
     <script>
       msg (Template("DefaultHelp"))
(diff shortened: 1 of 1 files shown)


WorldModel/WorldModel/Core/CoreCommands.aslx
```diff
@@ -913,35 +913,26 @@
...
     <pattern type="string">^save$</pattern>
     <script>
-      if (HasAttribute(game, "questplatform")) {
-        if (game.questplatform = "desktop") {
-           request(RequestSave, "")
-        }
-        else {
-          JS.saveGame ()
-        }
+      if (IsDesktop()){}
```
alexwarren, line 926: The open and close braces here looks like an error?

alexwarren (commented):
I'm not sure why we should have separate code for desktop and web within an ASLX file. Shouldn't `RequestSave` do the right thing instead?

WorldModel/WorldModel/Core/CoreCommands.aslx
KVonGit: > The open and close braces here looks like an error?

This is proof that I should switch back to Notepad++. I like VS Code, but I apparently need to "practice" with it before using it for important things.

I just fixed this.

KVonGit: > I'm not sure why we should have separate code for desktop and web within an ASLX file. Shouldn't `RequestSave` do the right thing instead?

I agree.

I looked at that code a few times before doing it this way, and I don't know what that C# code is doing. If you can make `RequestSave` handle it, that sounds much better than this.

alexwarren: Is there an Issue for this one?

KVonGit: 10-4

https://github.com/textadventures/quest/issues/1200

PR #1375, opened 2025-01-15
Add HTML element text processor commands

### LINKED ISSUE #1337: Add HTML element text processor commands (#feature-request #quest59)
We've already got the code for this, and it was in the last available testing build.

**CoreTypes** (add to `game.textprocessorcommands`)
```xml
      <item key=":">
        game.textprocessorcommandresult = ProcessTextCommand_Element (section, data)
      </item>
      <item key="/">
        game.textprocessorcommandresult = ProcessTextCommand_Element (section, data)
      </item>
    </textprocessorcommands>
```

**CoreOutput.xml**
```xml
  <function name="ProcessTextCommand_Element" parameters="section, data" type="string"><![CDATA[
    if (IsRegexMatch(":?/?\\w+\\b((?<=\\s)\\w+\\b|(?<=\\s)\\w+=\\w+\\b|(?<=\\s)\\w+=\"[^\"]*\"|(?<=\\s)\\w+='[^']*')*/?\\s*", section)) {
      if (StartsWith (section, ":")) section = LTrim (Mid (section, 2))
      return("<" + section + ">")
    }
    else {
      return ("@@@open@@@" + ProcessTextSection (section, data) + "@@@close@@@")
    }
  ]]></function>
```

---
image

---
https://textadventures.co.uk/forum/quest/topic/ykouw7idaeytupcl3rmywa/isnt-there-an-easy-way-to-add-text-processor-commands-solved

KVonGit:
Closes #1337

Allows things like this (for the WebEditor users, who can't always use `<`):
```
{:div id="test-div"}{:span id="test-span"}I am the text in the SPAN.{/span}{/div}
```

`<div id="test-div"><span id="test-span">I am the text in the SPAN.</span></div>`

### DIFF
--- a/WorldModel/WorldModel/Core/CoreOutput.aslx
+++ b/WorldModel/WorldModel/Core/CoreOutput.aslx
@@ -769,6 +769,16 @@
       }
     }
   ]]></function>
+
+  <function name="ProcessTextCommand_Element" parameters="section, data" type="string"><![CDATA[
+    if (IsRegexMatch(":?/?\\w+\\b((?<=\\s)\\w+\\b|(?<=\\s)\\w+=\\w+\\b|(?<=\\s)\\w+=\"[^\"]*\"|(?<=\\s)\\w+='[^']*')*/?\\s*", section)) {
+      if (StartsWith (section, ":")) section = LTrim (Mid (section, 2))
+      return("<" + section + ">")
+    }
+    else {
+      return ("@@@open@@@" + ProcessTextSection (section, data) + "@@@close@@@")
+    }
+  ]]></function>
   
   <function name="SetFramePicture" parameters="filename">
     <![CDATA[
--- a/WorldModel/WorldModel/Core/CoreTypes.aslx
+++ b/WorldModel/WorldModel/Core/CoreTypes.aslx
@@ -202,6 +202,12 @@
       <item key="=">
         game.textprocessorcommandresult = ProcessTextCommand_Eval (section, data)
       </item>
+      <item key=":">
+        game.textprocessorcommandresult = ProcessTextCommand_Element (section, data)
+      </item>
+      <item key="/">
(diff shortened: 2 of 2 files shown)


WorldModel/WorldModel/Core/CoreOutput.aslx
```diff
@@ -769,6 +769,16 @@
       }
     }
   ]]></function>
+
+  <function name="ProcessTextCommand_Element" parameters="section, data" type="string"><![CDATA[
+    if (IsRegexMatch(":?/?\\w+\\b((?<=\\s)\\w+\\b|(?<=\\s)\\w+=\\w+\\b|(?<=\\s)\\w+=\"[^\"]*\"|(?<=\\s)\\w+='[^']*')*/?\\s*", section)) {
```
alexwarren, line 774: That's... quite the Regex. Where did that come from and what is it supposed to do?

alexwarren (commented):
I'm a bit dubious of this one, if it's just for WebEditor users who are facing some bug when they use `<`, we should just fix that instead.

WorldModel/WorldModel/Core/CoreOutput.aslx
KVonGit: > I'm a bit dubious of this one, if it's just for WebEditor users who are facing some bug when they use `<`, we should just fix that instead.

That's all it's for. I thought not being able to use `<` in the WebEditor was a known issue with no workaround. If that can be fixed, that sounds much better.

---
> That's... quite the Regex. Where did that come from and what is it supposed to do?

Haha... It works. have no clue *how* it works, but it handled everything Pertex and I could think of to test it.

It came from mrangel in the forums: https://textadventures.co.uk/forum/quest/topic/ykouw7idaeytupcl3rmywa/isnt-there-an-easy-way-to-add-text-processor-commands-solved

PR #1395, opened 2025-01-20
Recreate using_templates.md

KVonGit:
- Copied from https://web.archive.org/web/20131014154837/http://quest5.net/wiki/Using_Templates and slightly modified for GH Pages

### DIFF
--- a/docs/using_templates.md
+++ b/docs/using_templates.md
@@ -1,0 +1,95 @@
+---
+layout: index
+title: Using Templates
+---
+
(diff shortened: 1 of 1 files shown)


KVonGit:
There's something about this one that feels off, but I can't put my finger on it.

I am probably imagining things, but here it is in my fork (where I have activated Pages for testing purposes):

https://kvongit.github.io/quest/using_templates.html

Issue #1447, opened 2025-02-22
Viva WebPlayer Can't Handle boolean ASLEvent


Pertex:
https://discord.com/channels/1278365322054996018/1339730685664952360/1342981897977987072

here is a bug we found with ASLEVENT. If you try to return a boolean, Viva reacts strange. Just start the game  and wait until you are told to insert a name. When you try to press 'return' nothing happens. I can solve that problem when i use 

 function testOnline() {   
    var check = "false";
    alert(window.location.hostname);
    if( window.location.hostname != "local") {
        check="true";
    } 
    ASLEvent("checkOnline", check);
    return false;
}
 
instead of 

function testOnline() {   
    var check = false;
    alert(window.location.hostname);
    if( window.location.hostname != "local") {
        check=true;
    } 
    ASLEvent("checkOnline", check);
    return false;
} 

in system.js 

[rbc.zip](https://github.com/user-attachments/files/18925856/rbc.zip)

alexwarren:
From looking at the code, I think I see the problem. The desktop player concatenates the event name and parameter, whereas WebPlayer v5 and Viva send these separately. That means in the desktop player, the parameter is always coerced into a string.

So the fix will be in the `ASLEvent` function:

```
await WebPlayer.uiSendEvent(event, "" + parameter);
```

and we should make the change both in the `v5` and `main` branches.

KVonGit:
It looks like this is already happening in V5:

[` $("#fldUIMsg").val("event " + event + ";" + parameter);`](https://github.com/textadventures/quest/blob/094f091f54cc5d26d9c3627a536db235725e03e2/WebPlayer/player.js#L175)

alexwarren:
> It looks like this is already happening in V5(?):
> 
> [` $("#fldUIMsg").val("event " + event + ";" + parameter);`](https://github.com/textadventures/quest/blob/094f091f54cc5d26d9c3627a536db235725e03e2/WebPlayer/player.js#L175)

Ah yes you're right, so we only need this change in Viva.

PR #1449, opened 2025-02-23
Coerce ASLEvent parameter into string

KVonGit:
- The `ASLEvent` parameter will be coerced into a string

See #1447


alexwarren:
Thanks! Have you tested this fixes the issue with Pertex's code (just don't want to rely entirely on my word for what I reckon the fix is)

KVonGit:
> Thanks! Have you tested this fixes the issue with Pertex's code (just don't want to rely entirely on my word for what I reckon the fix is)

My pleasure, and no I didn't test things with your code. :)

Just checked it though, and it fixed it.

Issue #1464, opened 2025-03-02
Using `* This is a note` is not a recognised command when multiple commands are allowed


KVonGit:
Example:

![Image](https://github.com/user-attachments/assets/e50ac11c-c707-426e-852e-be41190467ae)

---
I have a fix for this, and a fix for #1270 is in the same code (because that issue concerns the same function(s)).

PR on the way.

alexwarren:
@KVonGit Is this fixed now that #1473 is merged?

KVonGit:
![Image](https://github.com/user-attachments/assets/f56603c8-7a1c-4424-8ccc-6c907ab15251)

---
That's running an ASLX file in a fresh build of Viva, just after pulling all the available changes just now.

---
```xml
<asl version="580">

  <include ref="English.aslx"/>
  <include ref="Core.aslx"/>

  <game name="Multiple Commands for Viva">
    <gameid>57f6e69c-18c8-44d1-a8cd-9ce7bdd7b60c</gameid>
    <version>1.0</version>
    <firstpublished>2025</firstpublished>
  </game>

  <object name="room">
    <inherit name="editor_room" />

    <object name="player">
      <inherit name="editor_object" />
      <inherit name="editor_player" />
    </object>
  </object>

</asl>
```

[Multiple Commands for Viva.zip](https://github.com/user-attachments/files/19216423/Multiple.Commands.for.Viva.zip)

KVonGit:
Scratch that. I'm not always smart enough to save a file before a test, come to find out!

It works.

![Image](https://github.com/user-attachments/assets/ea2d2bc8-bd7b-41ba-bd04-517f98504da2)

Issue #1270, opened 2024-12-05
An alias including "." doesn't play well with `game.multiplecommands`


KVonGit:
If `game.multiplecommands` is `true` and there is a `.` in an object's alias, there will be issues.

> **\> x Mr. XanMag. Take beer. Give beer to Mr. Xanmag**
Nothing out of the ordinary.
I don't understand your command.
You pick it up.
Mr. XanMag thanks you kindly and gives you the **small brass key**.
I don't understand your command.

---
I have a fix based on the code posted by angelwedge in this issue: https://github.com/textadventures/quest/issues/1256#issuecomment-2515559421

I'll submit the PR after a little more testing.

angelwedge:
I'm not particularly happy with the solution I proposed there; as it only works with verb menus/buttons. Really it should work with typed commands as well.

Second solution would be checking object aliases for dots (during UpdateObjectLinks, maybe?) and making a list to check and escape dots in the same way; but that still only works if the name is exactly right.

My best idea for a solution (still not perfect) would be a later check. After generating a list of partial matches for an object name, you could check if multicommand mode is on, and there are commands remaining in the command queue, and the command string ends with the current string we're trying to match, and the alias contains the last word of the command string immediately followed by a dot. (could also have some additional checks; like the next command in the queue not matching any command regexes on its own; or the first word of the next command being in the possible alias). If so, stick the current and next command strings together with a dot, stick it on the beginning of the command queue, and fire HandleNextCommandQueueItem rather than proceding to the next name to resolve.

KVonGit:
Oh, I didn't even consider the links or the buttons. I only tested this one by typing the commands in a new test game -- with none of the stuff described in that other issue added to it.

You pointed out the flaw with `.` in an alias in [that other issue](https://github.com/textadventures/quest/issues/1256), but that only seemed like one thing out of many happening in that user's code, haha. So I created this issue by itself to make sure this bug gets addressed in the upcoming release.

Do you think I can/should change this to work with that? Or would that be completely different code?

This is the only function I have modified concerning this, and it seem to work during testing (but I do work at the Overlook Hotel):

```xml
  <function name="HandleCommand" parameters="command, metadata"><![CDATA[
    handled = false
    if (game.menucallback <> null) {
      if (HandleMenuTextResponse(command)) {
        handled = true
      }
      else {
        if (game.menuallowcancel) {
          ClearMenu
        }
        else {
          handled = true
        }
      }
    }
    if (not handled) {
[...]

PR #1465, opened 2025-03-02
Fix `*Note` (and alias with `.`) when multiple commands are allowed

KVonGit:
- Handle `*note` when `game.multiplecommands` is `true`
- Handle `.` in object alias when `game.multiplecommands` is `true`

See #1464 and #1270 -- to remain open until fixed in v5 branch as well


KVonGit:
There's still an issue with `.` an alias. Closing this PR and creating a new that only fixes the issue with the `*Note` command.

PR #1470, opened 2025-03-03
Fix CAS Resources in v5 WebPlayer

KVonGit:
- Use lower case filenames when reading game file from Azure blob to properly load resources
- Never listen to ALSTUNE ever again! (Sorry, Al!)


KVonGit:
Outdated.

(Plus, it works when the filenames are lower case.)

PR #1710, opened 2026-05-07
Fix map not showing room with no exits, and invisible exits being drawn

### LINKED ISSUE #1584: Map issues
If you are in a room without exits, the room is not drawn on the map.

If an exit is invisible, it will still be drawn on the map.

alexwarren:
Fixes #1584


Pertex:
:-) Do you need the 'if' in this function?

``` 
<function name="Grid_DrawPlayerInRoom" parameters="room">
    room.grid_render = true
    if (room.grid_render) { ....
```

alexwarren:
Good point!

PR #1935, opened 2026-07-25
feat(AppShell): hide empty advanced tree categories

alexwarren:
## Summary
- Workstream 2 of `docs/editor-progressive-disclosure.md`: the Functions, Timers, Walkthrough, and Advanced-group (Included Libraries, Templates, Dynamic Templates, Object Types) tree headers are always present in the underlying tree data (added unconditionally by `EditorController.AddTreeHeader`), even with nothing under them yet. Empty headers are noise, so `TreePanel.svelte` now prunes them from the rendered tree while empty.
- "Objects" is never hidden. "Javascript" (also a permanent child of "Advanced") is intentionally out of scope per the doc.
- Adding the first element of a hidden category via the Toolbar "+" menu (already header-independent) makes its header reappear with the new element visible.
- Pure Svelte-layer change — no EditorCore tree-event changes.

## Test plan
- [x] `npm run check` (svelte-check) clean
- [x] New Playwright script `tests/e2e/verify-appshell-empty-advanced-categories.mjs`: Functions/Timers/Walkthrough hidden on a fresh game; "Advanced" itself stays visible (default Included Libraries aren't empty) while Templates/Dynamic Templates/Object Types inside it are hidden; Objects never hidden; adding a Function via the Toolbar "+" menu makes "Functions" reappear with the new element, other empty categories stay hidden
- [x] Verified filter-textbox interaction manually — filtering already prunes empty branches correctly, no special handling needed
- [x] Re-ran `verify-appshell-tap-same-tree-node.mjs`, `verify-appshell-tree-dropdown-position.mjs`, `verify-appshell-tree-filter.mjs`, `verify-appshell-tree-splitter.mjs` — no regressions

### DIFF
--- a/src/AppShell/src/components/TreePanel.svelte
+++ b/src/AppShell/src/components/TreePanel.svelte
@@ -57,7 +57,7 @@
     // depend on the header being visible. "_objects" is never hidden.
     const HIDE_WHEN_EMPTY = new Set([
         "_functions", "_timers", "_walkthrough",
-        "_include", "_template", "_dynamictemplate", "_objecttype",
+        "_include", "_template", "_dynamictemplate", "_objecttype", "_javascript",
         "_advanced",
     ]);
 
--- a/src/AppShell/src/components/TreePanel.svelte
+++ b/src/AppShell/src/components/TreePanel.svelte
@@ -47,6 +47,31 @@
         return (byParent.get(null) ?? []).map(build);
     }
 
+    // ── Empty advanced categories ────────────────────────────────────────────────
+
+    // Advanced category headers are always present in the tree data (added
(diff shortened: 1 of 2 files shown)


alexwarren:
Follow-up: "_javascript" is also a permanent child of "_advanced" (`EditorController.AddTreeHeader`, same as Included Libraries/Templates/Dynamic Templates/Object Types) but was left out of the doc's enumerated hide-list — pretty clearly an oversight rather than intentional, since the Toolbar "+" menu already has a header-independent "Add JavaScript" entry, same pattern as the others. Added it to `HIDE_WHEN_EMPTY` for consistency and extended the verify script to cover it.

PR #1936, opened 2026-07-25
feat(AppShell): demote advanced categories in the script adder

alexwarren:
## Summary
- Workstream 3 of `docs/editor-progressive-disclosure.md`: script command categories where *every* visible command is `<advanced/>` (mirroring v5's `GetCategories` filter in reverse) now sort below a divider in `AddScriptModal`'s category sidebar, instead of appearing as peers of everyday categories like "Output".
- `WasmEditorBridge.ScriptCategoryInfo` gains `Advanced`, computed per category from `EditableScriptData.IsVisibleInSimpleMode`.
- Every command stays reachable — the filter box searches all categories regardless of the divider.

## Test plan
- [x] `dotnet build` (Release) for WasmEditor
- [x] `npm run check` (svelte-check) clean
- [x] New Playwright script `tests/e2e/verify-appshell-scriptadder-advanced-categories.mjs`: enables the (off-by-default) "Lightness and darkness" feature so the entirely-advanced "Darkness" category has visible commands, confirms it sorts after non-advanced categories with a divider between them, confirms its commands are still addable, and confirms the filter box still surfaces them regardless of the divider
- [x] Re-ran `verify-appshell-addscript-modal-mobile.mjs`, `verify-appshell-script-adder-filter.mjs`, `verify-appshell-script-label-stacks-below.mjs`, `verify-appshell-scripteditor-row-buttons.mjs`, `verify-appshell-scripteditor-selection-toolbar.mjs` — no regressions

### DIFF
--- a/src/AppShell/src/components/AddScriptModal.svelte
+++ b/src/AppShell/src/components/AddScriptModal.svelte
@@ -54,6 +54,15 @@
 
     const selectedCategory = $derived(categories[selectedCategoryIndex] ?? null);
 
+    // Within a mixed category (not already-entirely-advanced, which already got its own
+    // category-level divider above), commands are pre-sorted non-advanced-first by the
+    // bridge — find where the advanced tail starts so a divider can mark it.
+    const firstAdvancedCommandIndex = $derived(
+        selectedCategory && !selectedCategory.advanced
+            ? selectedCategory.commands.findIndex((c) => c.advanced)
+            : -1
+    );
+
     // ── Filtering ──────────────────────────────────────────────────────────────
     // While the box has text, the category sidebar + single-category list is
     // replaced by one flat list of matches across every category, each tagged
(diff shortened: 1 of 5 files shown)


alexwarren:
Pushed a follow-up: category-level demotion alone is invisible on a stock game (the only entirely-advanced categories, Darkness and Drawing, are both gated behind off-by-default features), so there was no way to see this PR's effect without flipping a feature flag first.

Extended the same idea one level down — within a mixed category, advanced commands now sort after a plain "Advanced" divider too (same non-collapsing treatment as the category sidebar, so the modal isn't mixing two different demotion metaphors). E.g. opening "Scripts" now shows If.../First time above the divider, Comment/For/For each/Switch/While/etc. below it — visible on any fresh game, no setup required.

Also fixed the existing category-level test: main pulled in #1934 since I opened this PR, so "Lightness and darkness" (itself `<advanced/>`) is now folded into the property editor's own Advanced expander and needed that opened first.

PR #2028, opened 2026-08-10
feat(AppShell): new german language file

Pertex:


alexwarren:
Thank you for kicking off the German translation! This covered PR1+PR2 of the AppShell i18n rollout (~155 keys). Since then, PR3–PR5 landed and added ~206 more keys (attribute/script editors, banners/modals, and the Home/Play/Create routes), so this file is now missing a good chunk of the current UI.

I've taken your translations as the base, completed the remaining keys, and fixed a few small issues found along the way (a couple of untranslated strings, a copy-paste bug in one of the delete-confirmation messages, and a typo) — see #2030. Closing this one in favor of that, but wanted to flag it here first and credit the head start. Thanks again for the contribution!

alexwarren:
(Claude wrote that previous comment, I haven't suddenly gained the ability to write German. But thanks for contributing that @Pertex !)

PR #2032, opened 2026-08-11
feat: add German gamebook template and editable language field

alexwarren:
## Summary

- Gamebooks previously had exactly one creation template — English captions baked into `GamebookCore.aslx`'s hardcoded `EditorEnglish.aslx` include, and no `<languageid>` at all — even though a full German translation of the gamebook-specific editor captions (`EditorDeutsch.aslx`, all 38 `EditorGB*` keys) already existed and was simply never wired up. Text Adventures already support this via a per-language template (`English.template`, `Deutsch.template`, ...), each pairing an `Editor*.aslx` caption set with a `<languageid>` value.
- Adds the same picker for Gamebooks: a new `Gamebook-Deutsch.template` alongside the existing `Gamebook.template` (which now explicitly pairs with `EditorEnglish.aslx`/`languageid=en` instead of `GamebookCore.aslx` hardcoding the include), `GamebookCore.aslx`'s `defaultgame` type now sets `<languageid>` (matching `CoreTypes.aslx`'s Text Adventure equivalent), and the AppShell template picker (`open/+page.svelte`) shows the same `<select>` for Gamebooks it already shows for Text Adventures.
- `game.languageid` is now exposed as an editable Properties field (dropdown + freetext), so an author can retag it to any language after creation — this covers "I'm writing in some other language" without needing a dedicated `Editor*.aslx` translation for every language.

## Also fixes along the way

- `EditorController.GetAvailableTemplates()` keyed its dictionary by display label, which throws once two templates share a label across game types (e.g. "Deutsch" for both the Text Adventure and Gamebook templates). Now keyed by the already-unique resource name.
- `Engine.csproj` lists embedded template resources explicitly rather than via a glob — the new template needed an explicit entry, or it would silently never be served (caught this because the AppShell dropdown was still showing only one option after adding the `.template` file).

## Test plan

- [x] `dotnet build --configuration Release`
- [x] `dotnet test` (full suite, 337 tests)
- [x] `npm run check` / `npm run lint` in `src/AppShell`
- [x] Manual verification in the AppShell dev server: German and English gamebook creation, German captions + translated starter page text, `languageid` field defaulting correctly and accepting freetext codes
- [x] New `tests/e2e/verify-appshell-gamebook-language-picker.mjs`, checked in per repo convention, covering both language paths and the freetext "Other" case

### DIFF
--- a/src/AppShell/src/routes/open/+page.svelte
+++ b/src/AppShell/src/routes/open/+page.svelte
@@ -169,10 +169,13 @@
     let textAdventureTemplates = $derived(templates.filter(tpl => tpl.type === "textadventure"));
     let gamebookTemplates = $derived(templates.filter(tpl => tpl.type === "gamebook"));
     let lastTextAdventureTemplateId = $state("");
+    let lastGamebookTemplateId = $state("");
 
     $effect(() => {
         if (selectedTemplate?.type === "textadventure") {
             lastTextAdventureTemplateId = selectedTemplate.id;
+        } else if (selectedTemplate?.type === "gamebook") {
+            lastGamebookTemplateId = selectedTemplate.id;
         }
     });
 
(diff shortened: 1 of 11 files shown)


alexwarren:
Found and fixed a backward-compatibility regression in the first commit: reopening a **pre-existing** gamebook (created before this PR) showed unresolved `[EditorGBPageType]`-style placeholders throughout the Properties panel, because `GamebookCore.aslx` no longer supplied any editor captions on its own, and old game files never had their own top-level `Editor*.aslx` include (that responsibility used to live inside `GamebookCore.aslx`, which is loaded live while editing, not frozen).

Second commit restores that baseline and fixes the override mechanism properly — `<caption>` text is template-substituted *eagerly at XML parse time*, not lazily when the editor UI renders it, so simply reordering `<include>`s doesn't let a later `EditorDeutsch.aslx` override captions already resolved by `GamebookCoreEditor.aslx`. The fix routes the choice of which `Editor*.aslx` to include through a template itself (`GamebookEditorLibrary`), which a per-language template can override via `GameLoader`'s preliminary base-file scan — same mechanism `LanguageId` already relied on. `Gamebook.template` (English) is now byte-identical to its pre-existing form, and a new e2e regression case imports a raw pre-existing-shaped gamebook file to guard against this specific failure mode going forward.

PR #2134, opened 2026-08-24
fix(Engine): announce game output and label controls for screen readers

alexwarren:
## Summary
- Adds `role="log" aria-live="polite"` to `#divOutput` (the game transcript, shared by WebPlayer and WasmPlayer) so new turn text is announced by screen readers as it appears.
- Labels the command input, hamburger menu, compass buttons, and save-name inputs, which previously had no accessible name (or relied on `title` alone).
- Wires `aria-labelledby` on the Save/Load and Debugger dialogs in both WebPlayer (Razor) and WasmPlayer.
- Adds two new `ChromeStrings` keys (`CommandInputLabel`, `ShowPanesButton`) following the existing per-language localization convention.

First of a planned series of accessibility PRs covering both the player and the AppShell editor.

## Test plan
- [x] `dotnet build --configuration Release` — succeeds
- [x] `dotnet test --configuration Release` — all 384 tests pass
- [x] Verified in a WasmPlayer dev build via the accessibility tree: `#divOutput` reports `role="log"`/`aria-live="polite"`, typing a command appends new narrative text into that same live region, and the command input/hamburger/compass buttons/save dialogs all report the expected accessible names.

### DIFF
--- a/src/Engine/Core/Languages/Deutsch.aslx
+++ b/src/Engine/Core/Languages/Deutsch.aslx
@@ -582,6 +582,8 @@
   <template name="RestartingBusy">Startet neu…</template>
   <template name="DeleteSaveConfirm">„[name]“ löschen? Das kann nicht rückgängig gemacht werden.</template>
   <template name="SavedGameDefaultLabel">Spielstand — [timestamp]</template>
+  <template name="CommandInputLabel">Gib deinen Befehl ein</template>
+  <template name="ShowPanesButton">Bereiche anzeigen</template>
 
   <!-- Explicit copies of English.aslx's own values (both already reached via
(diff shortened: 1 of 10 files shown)


alexwarren:
Follow-up pushed: `CommandInputLabel`/`ShowPanesButton` were only added to `ChromeStrings.cs`'s C# `Defaults` dict, not as `<template>` entries in `Engine/Core/Languages/*.aslx` like every other ChromeStrings key — so they always fell back to English regardless of language, and `tools/i18n/audit.mjs` had no visibility into them at all (its baseline is derived from what's declared in the language files). Added to `English.aslx` plus German/Spanish translations.

PR #2166, opened 2026-08-29
test(e2e): fix stale "GAME OBJECTS"/"PROPERTIES" pane-header locators

alexwarren:
## Summary
- The nightly `e2e` workflow on `main` was failing (`appshell_chromium` and `electron` jobs both stuck on `waiting for locator('text=GAME OBJECTS') to be visible`).
- Root cause: #2144 removed TreePanel's static "Game Objects"/"Game Pages" header and PropertyEditor's static "Properties" fallback header as redundant, and updated 3 e2e scripts that referenced that text — but missed 4 more.
- Swaps the stale `text=GAME OBJECTS` / `text=PROPERTIES` locators for stable substitutes already used elsewhere in the suite: the tree pane's always-present filter input (`input[placeholder="Filter..."]`) as the "tree pane active" marker, and the properties pane's "Setup" tab button as the "properties pane active" marker. `verify-electron-undo-textarea.mjs`'s "blur to the tree" step (which clicked the removed header) now calls `document.activeElement.blur()` directly instead of relying on a click target.

While driving this to a real green CI run, found and fixed three more issues surfaced only by actually running the full suite (not visible from a diff read):

- `verify-appshell-i18n-pseudoloc.mjs`'s `button[aria-label]` locator for TreePanel's view-options menu was similarly stale: #2139 added `aria-label` to Toolbar's Back/Forward/BackToHome buttons, which render earlier in the DOM and start disabled — scoped to `#workspace-content` (added by #2141) to disambiguate.
- The same script's `AttributesEditor` header-count assumption ("index 0 is PropertyEditor's own Properties header") broke for the same reason as the original fix — adjusted the expected indices/count.
- The same script's library-verb visibility check (waiting on "drink" from `CoreCommands.aslx`) was hitting a second, deeper level of tree collapsing: `TreePanel.svelte`'s `groupLibraryChildren` folds 20+ consecutive same-library verbs into their own nested collapsed group, so "drink" sits two levels below "Verbs", not one. Now uses the tree node's own "Expand all below" context action instead of a single expand click, so it doesn't depend on the exact nesting depth.
- Discovered (via real rate-limit failures during the above re-runs) that most `verify-appshell-*.mjs` scripts fire an unauthenticated request to `https://api.github.com/repos/textadventures/quest/releases/latest` as a side effect of `HomeHeader`/`DownloadButton` mounting on `/` or `/open` whenever `PUBLIC_SHOW_HOME` is set (which the `appshell_chromium` job's dev server always is) — easily exhausting GitHub's 60-req/hour unauthenticated limit across a full suite run. Added a default mock for that endpoint in `tests/e2e/lib/tracked-chromium.mjs`, the shared `chromium` wrapper every one of those scripts already imports instead of `playwright` directly, so the whole suite is now immune to it. Scripts with their own intentional mocks (including a failure-path test) are unaffected, since Playwright always prefers a page-level `page.route()` over this context-level default.

## Test plan
- [x] `node --check` on every edited script
- [x] `node tests/e2e/check-workflow-manifest.mjs` passes
- [x] Full `e2e` workflow run ([33253531267](https://github.com/textadventures/quest/actions/runs/33253531267)) — all 8 jobs green, including `appshell_chromium` and `electron`

https://claude.ai/code/session_01TUYjsmoYFAYgTkLtQAhmVX

---
_Generated by [Claude Code](https://claude.ai/code/session_01TUYjsmoYFAYgTkLtQAhmVX)_

### DIFF
--- a/tests/e2e/lib/tracked-chromium.mjs
+++ b/tests/e2e/lib/tracked-chromium.mjs
@@ -1,7 +1,10 @@
 // Drop-in replacement for `import { chromium } from 'playwright'`, used by every
-// verify-appshell-*.mjs script instead of importing 'playwright' directly. When
-// QUEST_TOUCHED_FILES_DIR is unset (every normal run, local or CI) this is a pure
-// passthrough to the real `chromium` export with zero behavioral difference.
+// verify-appshell-*.mjs script instead of importing 'playwright' directly.
+//
+// Always mocks GitHub's releases API (see mockReleasesApi below) on every
+// context/page a launched browser creates. Beyond that, when
+// QUEST_TOUCHED_FILES_DIR is unset (every normal run, local or CI) this adds no
+// other behavior over the real `chromium` export.
 //
 // When QUEST_TOUCHED_FILES_DIR is set, records real V8 JS coverage for every page
 // a launched browser opens and, on browser.close(), writes the set of AppShell
@@ -39,7 +42,62 @@ import { join, basename } from 'node:path';
 const outDir = process.env.QUEST_TOUCHED_FILES_DIR;
 const SNAPSHOT_INTERVAL_MS = 4000;
 
-export const chromium = outDir ? withTracking(realChromium, outDir) : realChromium;
+// HomeHeader mounts DownloadButton (src/AppShell/src/lib/download-links.ts)
+// on every page load when PUBLIC_SHOW_HOME is set (every appshell_chromium
+// e2e script does, via the workflow's dev-server env), which fires this
+// exact request unauthenticated. Almost every verify-appshell-*.mjs script
+// navigates to / or /open at least once, so left unmocked here this was
+// hitting GitHub's 60-req/hour unauthenticated rate limit across a full
+// suite run - default-mocked for every context/page so individual scripts
+// don't each need their own copy. A script that wants specific release data
+// (or to test the failure path) can still override with its own page.route()
+// for the same URL - Playwright always prefers a page-level route over this
+// context-level one, regardless of registration order.
+const RELEASES_API_URL = 'https://api.github.com/repos/textadventures/quest/releases/latest';
+const SAMPLE_RELEASE = {
+    tag_name: 'v6.0.0-beta.42',
+    published_at: '2026-07-19T07:30:55Z',
+    assets: [
+        { name: 'Quest.Viva.Setup.6.0.0-beta.42.exe', browser_download_url: 'https://example.com/win.exe' },
+        { name: 'Quest.Viva-6.0.0-beta.42-arm64.dmg', browser_download_url: 'https://example.com/mac.dmg' },
+        { name: 'Quest.Viva-6.0.0-beta.42.deb', browser_download_url: 'https://example.com/linux.deb' },
+        { name: 'Quest.Viva-6.0.0-beta.42.AppImage', browser_download_url: 'https://example.com/linux.AppImage' },
+    ],
+};
+
+async function mockReleasesApi(context) {
+    await context.route(RELEASES_API_URL, route => route.fulfill({
+        status: 200,
+        contentType: 'application/json',
+        body: JSON.stringify(SAMPLE_RELEASE),
+    }));
+}
+
+function withDefaultMocks(real) {
+    return {
+        ...real,
+        async launch(...args) {
+            const browser = await real.launch(...args);
+            for (const ctx of browser.contexts()) await mockReleasesApi(ctx);
+            const origNewContext = browser.newContext.bind(browser);
+            browser.newContext = async (...a) => {
+                const ctx = await origNewContext(...a);
+                await mockReleasesApi(ctx);
+                return ctx;
+            };
+            const origNewPage = browser.newPage.bind(browser);
+            browser.newPage = async (...a) => {
+                const page = await origNewPage(...a);
+                await mockReleasesApi(page.context());
+                return page;
+            };
+            return browser;
+        },
+    };
+}
+
+const withDefaults = withDefaultMocks(realChromium);
+export const chromium = outDir ? withTracking(withDefaults, outDir) : withDefaults;
 
 function withTracking(real, outDir) {
     return {
--- a/tests/e2e/verify-appshell-assetpicker-mobile-overflow.mjs
+++ b/tests/e2e/verify-appshell-assetpicker-mobile-overflow.mjs
@@ -25,8 +25,7 @@ try {
     await mobilePage.click('text=room');
     await mobilePage.waitForSelector('button:has-text("room")', { timeout: 5000 });
     await mobilePage.click('button:has-text("room")'); // back to tree
-    await mobilePage.waitForSelector('text=GAME OBJECTS', { timeout: 5000 });
-    // Exact match — "text=game" (substring) also matches the "GAME OBJECTS" heading.
+    await mobilePage.waitForSelector('input[placeholder="Filter..."]', { timeout: 5000 });
     await mobilePage.click('text="game"');
     await mobilePage.waitForSelector('button:has-text("Setup")', { timeout: 5000 });
 
--- a/tests/e2e/verify-appshell-i18n-pseudoloc.mjs
+++ b/tests/e2e/verify-appshell-i18n-pseudoloc.mjs
@@ -55,27 +55,8 @@ const baseUrl = process.argv[2] || 'http://localhost:5180';
 const browser = await chromium.launch();
 const page = await browser.newPage({ viewport: { width: 1280, height: 900 } });
 page.on('pageerror', err => console.log('[pageerror]', err.message));
-
-// DownloadButton (checked below) hits GitHub's real releases API
-// unauthenticated (src/AppShell/src/lib/download-links.ts) — mocked here so
-// this script doesn't depend on network access or GitHub's unauthenticated
-// rate limit (60 req/hour/IP, easily exhausted by repeated e2e runs from the
-// same CI runner). Mirrors the sample release used in
-// verify-appshell-download-button.mjs.
-await page.route('https://api.github.com/repos/textadventures/quest/releases/latest', route => route.fulfill({
-    status: 200,
-    contentType: 'application/json',
-    body: JSON.stringify({
-        tag_name: 'v6.0.0-beta.42',
-        published_at: '2026-07-19T07:30:55Z',
-        assets: [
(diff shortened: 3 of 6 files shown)


alexwarren:
The manual `e2e` run ([33240627639](https://github.com/textadventures/quest/actions/runs/33240627639)) shows the `electron` job red, but it's not this PR's failure — it died before any test body ran:

```
Downloading Electron binary...
TypeError: fetch failed
    at node:internal/deps/undici/undici:15141:13
    ...
Error: Electron failed to install correctly. Please delete `node_modules/electron` and run "npx install-electron --no" manually.
    ...
    at file:///Users/runner/work/quest/quest/tests/e2e/verify-electron-open-file.mjs:27:83
```

That's `verify-electron-open-file.mjs` (a script this PR doesn't touch) failing to even launch Electron — a network hiccup fetching the Electron binary on the macOS runner, not a code issue. Because the job fails fast, `verify-electron-selectall-textbox.mjs` and `verify-electron-undo-textarea.mjs` (the two scripts this PR actually fixes) never got a chance to run.

I don't have permission to re-run the job myself (403 on `rerun_failed_jobs`) — could someone with write access re-run just the failed `electron` job? Meanwhile `appshell_chromium` is green so far, including `verify-appshell-assetpicker-mobile-overflow.mjs` (one of the fixed scripts) passing.

---
_Generated by [Claude Code](https://claude.ai/code)_

alexwarren:
Update on that same run ([33240627639](https://github.com/textadventures/quest/actions/runs/33240627639)): `appshell_chromium` also went red, but **not from this PR's changes** — both scripts this PR touches passed cleanly:
- `verify-appshell-assetpicker-mobile-overflow.mjs` — step 22, PASS
- `verify-appshell-mobile-master-detail.mjs` — step 37, PASS

The job later failed at step 61, `verify-appshell-i18n-pseudoloc.mjs`, in code neither this PR nor its scripts touch:

```
PASS: [TreePanel view-options title] "[Báck]"
FAIL: locator.click: Timeout 30000ms exceeded.
  - locator resolved to <button disabled ... title="[Báck]" aria-label="[Báck]" class="toolbar-icon-btn ...">
    - element is not enabled
```

Root cause: `verify-appshell-i18n-pseudoloc.mjs:227` assumes `page.locator('button[aria-label]').first()` uniquely matches TreePanel's view-options button. #2139 ("keyboard-operable dropdown menu and icon button labels") added `aria-label` to Toolbar's Back/Forward/BackToHome buttons — which render earlier in the DOM and are disabled at load (no history yet) — so `.first()` now grabs the disabled Toolbar Back button instead. The `assertPseudo` check on its title only verifies the string is bracket-wrapped, not that it's the *right* string, so it silently "passed" on the wrong element before failing on the subsequent `.click()`.

This is the same class of stale-locator regression this PR fixes, just a different pair of PRs (#2139 and #2144, both landed close together) and a different file — not something I should fold into this PR's diff. A scoped fix would be to anchor the locator to TreePanel's own filter row instead of a bare `button[aria-label]`, e.g.:

```js
const viewOptionsBtn = page.locator('input[aria-label]').locator('xpath=ancestor::div[contains(@class,"border-b")][1]').locator('button[aria-label]');
```

(same idea as the toast-locator fix in dbc2bd6.) Flagging here rather than pushing it — happy to open a follow-up PR for it if wanted.

---
_Generated by [Claude Code](https://claude.ai/code)_

alexwarren:
Pushed ca80333, folding the `verify-appshell-i18n-pseudoloc.mjs` fix from the previous comment into this PR (per request): both call sites of the unscoped `button[aria-label]').first()` now scope to `#workspace-content` (added by #2141), so they no longer collide with Toolbar's newly-aria-labelled Back/Forward/BackToHome buttons.

Still waiting on a re-run of the `electron` job (unrelated infra flake on the previous manual run — see the earlier comment) to confirm `verify-electron-selectall-textbox.mjs` and `verify-electron-undo-textarea.mjs`.

---
_Generated by [Claude Code](https://claude.ai/code)_

alexwarren:
Re-run of run [33241243805](https://github.com/textadventures/quest/actions/runs/33241243805) on ca80333 confirmed the aria-label fix worked (`[TreePanel view-options title] "[Tréé víéw óptíóns]"` — correct element now) and got the script much further, but it then hit a *second*, different assertion further down in the same script:

```
FAIL: AttributesEditor: expected 3+ headers (incl. PropertyEditor's), found 2
```

Same root cause, different assertion: the script's own comment said "index 0 is PropertyEditor's own 'Properties' header (same span classes) — AttributesEditor's own headers start at 1", but #2144 removed that static header too, so AttributesEditor's "Inherited types"/"Attributes" headers now start at index 0. Pushed 7cb998e fixing the index/count assumption.

All three fixed scripts (`assetpicker-mobile-overflow`, `mobile-master-detail`, `electron`'s `selectall-textbox`/`undo-textarea`) plus the earlier aria-label fix in `i18n-pseudoloc` are confirmed passing in real CI at this point — only this newly-discovered header-count assumption was still red. Could someone re-run the `e2e` workflow once more on the latest commit to confirm this one too?

---
_Generated by [Claude Code](https://claude.ai/code)_

alexwarren:
Third fix pushed (fea8c2a). Diagnosed the "drink" library-verb timeout from run [33241773974](https://github.com/textadventures/quest/actions/runs/33241773974): the failing run's own log showed the Show Library Elements toggle *and* the view-options click both working correctly (the PASS lines right before it), and the failure was `locator resolved to <span>drink</span> ... element is not visible` — so the "drink" library-verb node existed, just inside a collapsed ancestor.

Root cause: the script's `VerbsEditor` section (just above, in the same run) adds a "juggle" verb to `room`, which — per that section's own log line "the novel verb also created a new command element under the Verbs tree node" — plants a new child under the same global "Vérbs" tree branch. That likely auto-expands it, invalidating the old "Verbs starts collapsed" assumption baked into this step: the unconditional expand-click could be *re-collapsing* an already-expanded branch instead of opening it. Now it only clicks the toggle when "drink" isn't already visible. Also scoped both `Show Library Elements` menu-item interactions to `[role="menu"]` (unique to `DropdownMenu.svelte`) instead of a bare `.absolute` class, which by this point in the flow can also match unrelated hover-action overlays in other mounted components.

I tried re-running the workflow myself but still don't have permission (403). Could you kick off the `e2e` workflow once more on the latest commit?

---
_Generated by [Claude Code](https://claude.ai/code)_

alexwarren:
Re-run on fea8c2a ([33247812644](https://github.com/textadventures/quest/actions/runs/33247812644)) showed the previous fix didn't actually work — identical failure, "drink" resolving but still not visible, even though the Show Library Elements toggle was confirmed correct right before it.

Found the real root cause by checking source rather than guessing further: `src/Engine/Core/CoreCommands.aslx` defines 20+ consecutive `<verb>` elements (`speakto`, `buy`, `climb`, `drink`, `eat`, …). `TreePanel.svelte`'s `groupLibraryChildren` folds any run of 2+ same-library children into a nested, separately-collapsed `"librarygroup"` node — so once library elements are shown, `drink` sits **two** collapsed levels deep (Verbs → CoreCommands.aslx group → drink), not one. My previous fix only ever expanded the outer Verbs level.

Pushed 9abb1ab: instead of guessing expand-click counts/paths, it now uses the tree node's own "Expand all below" context-menu action (same right-click-via-`.node-actions`-button technique the script already uses for the `room` node's own context menu a bit earlier), which recursively expands every descendant regardless of nesting depth — so it doesn't matter how many levels are actually collapsed.

Could you kick off the `e2e` workflow once more on the latest commit?

---
_Generated by [Claude Code](https://claude.ai/code)_

alexwarren:
Good news / bad news on run [33251123503](https://github.com/textadventures/quest/actions/runs/33251123503) (commit 9abb1ab):

**Good:** the "drink" library-verb failure is gone — the script got past that whole section this time with no error there at all.

**Bad:** it died earlier instead, at `DownloadButton`:
```
FAIL: page.textContent: Timeout 30000ms exceeded.
  - waiting for locator('.download-dropdown .absolute p')
```
This is unrelated to this PR or any of my fixes. `src/AppShell/src/lib/download-links.ts` calls `https://api.github.com/repos/textadventures/quest/releases/latest` **unauthenticated**, and the log for this run shows repeated `[console.error] Failed to load resource: the server responded with a status of 403 ()` scattered across several unrelated test scripts (`exits-editor`, `verbs-editor`, `tree-splitter`, `unlock-exit-picker`, `rename-invalid-name`, `dialogue-pages`) — classic GitHub REST API rate-limit exhaustion (60 req/hr unauthenticated per IP). I've now had six manual `e2e` runs triggered today, each hitting that same endpoint from the runner's shared IP — almost certainly what exhausted it.

Since it's a rolling window, re-running immediately would likely hit the same 403. I'd suggest waiting ~30–60 minutes before the next run so the limit clears. I still can't confirm the `drink`/Verbs fix reached the finish line on this run since it died before getting back around to it in a full pass — though it did pass that section cleanly this time, which is a strong signal. I'll check back later once the rate limit should have reset.

---
_Generated by [Claude Code](https://claude.ai/code)_

alexwarren:
Went ahead and fixed the rate-limit flake for real (pushed 2857fb5 then a7544d5, after the user confirmed they wanted the broader fix):

- **2857fb5**: mocked GitHub's releases API in `verify-appshell-i18n-pseudoloc.mjs` specifically (the script that had actually failed).
- **a7544d5**: turned out that was too narrow — `HomeHeader` mounts `DownloadButton` (and fires this same unauthenticated `api.github.com/repos/textadventures/quest/releases/latest` call) on *every* page load once `PUBLIC_SHOW_HOME` is set, which the `appshell_chromium` job's dev server always is. Nearly every `verify-appshell-*.mjs` script navigates to `/` or `/open` at least once, so a full suite run was burning most/all of GitHub's 60-req/hour unauthenticated budget on this alone — the real cause of the 403s scattered through today's re-runs. Moved the mock into `tests/e2e/lib/tracked-chromium.mjs` (the shared `chromium` wrapper every one of those scripts already imports instead of `playwright` directly), registered on every context/page a launched browser creates, and dropped the now-redundant per-script version from the first commit. `verify-appshell-download-button.mjs`/`verify-site-download-page.mjs`, which already have their own intentional mocks (including a failure-path test), are unaffected — Playwright always prefers a page-level `page.route()` over this context-level default regardless of registration order.

This should make the whole suite immune to this particular flake going forward, not just today's failure. Could you kick off the `e2e` workflow once more on the latest commit?

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #2186, opened 2026-09-04
Add `<link rel=preload>` to start downloading subresources ASAP


dfabulich:
Load https://play.questviva.com/player/?id=yga06kzqtku8cn8vfgjt2a in Chrome Dev Tools Network tab with "Disable Cache" checked.

Resources are being loaded in "waterfalls", where one client-side request has to succeed before the next request can begin.

For example, in order to load https://playtextadventures.com/games/YGa06kzqtku8cN8vfGJt2A/1/resources/game.aslx we first have to request `wasm-player.js`, which then has to fetch `https://textadventures.co.uk/api/game/yga06kzqtku8cn8vfgjt2a?source=web&version=6.0.0-beta.56`, and only then does the browser know the `resourceRoot` to construct the URL to `game.aslx`.

Instead, `/player` should query the textadventures API on the server side. `resourceRoot` should be included as a global value in the initial HTML response, and `/player` should include [`<link rel=preload>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/preload) elements in the `<head>`.

That way, the browser can skip querying the textadventures API and just begin to download `game.aslx` ASAP, in parallel with `wasm-player.js`. When `wasm-player.js` wakes up and requests `game.aslx`, it may already be done fetching (especially if `game.aslx` is already cached).

Even more important is `dotnet.js` and the hundreds of WASM files it requests. All of those URLs (including hashes, see #2185) can be computed ahead of time, and added as `<link rel=preload>` elements.

Viva should start downloading those WASM files ASAP.

PR #2190, opened 2026-09-05
perf(WasmPlayer): cache static assets immutably instead of revalidating ~210 per load

### LINKED ISSUE #2185: Refreshing on the web should be much faster
Load a game in Quest Viva on the web, open Dev Tools (make sure "Disable Cache" is unchecked) and refresh the page.

Actual: It makes 214 requests when requesting the page. Most of them return 304 Not Modified, with `Cache-Control: public, max-age=0, must-revalidate`, forcing the browser to do network I/O to revalidate them.

Expected: It should refresh the main HTML document `/player`, and then it should skip requesting all of the subresources (JS, CSS, fetches, WASM) because they should be locally cached.

Specifically, `/player` should continue to use `Cache-Control: public, max-age=0`, but it should include a hash or other version number in a URL parameter for each subresource.

For example, instead of requesting https://play.questviva.com/player/quest-config.js, it should request `https://play.questviva.com/player/quest-config.js?hash=1abcdef`. When requesting the JS with that hash, it should use `Cache-Control: public, max-age=31536000, immutable`.

When `quest-config.js` updates, as the user reloads the `/player` page, they'll always request the root HTML document fresh. `/player` will request `quest-config.js` with its new hash e.g., `https://play.questviva.com/player/quest-config.js?hash=2abcdef`. Players who refresh will get the new `/player` page, but all (or almost all) of their hashes will be fresh, allowing them to skip all of those downloads entirely.

This is already correctly set up for the game's own resources; the URL includes the version number of the game itself, and so the resources are being correctly set to immutable.

This is especially important for `dotnet.js`, which requests hundreds of WASM files. If those WASM requests had hashes, those hundreds of requests could be skipped.

alexwarren:
Fixes #2185.

## The problem

Refreshing a game on play.questviva.com (cache **enabled**) makes ~213 requests, ~203 of which come back `304 Not Modified` to transfer about 90 KB in total. Measured against the live site: ~870 ms of pure revalidation on a fast connection, ~2.1 s at +100 ms RTT.

Three independent causes, and fixing any one of them alone changes nothing:

1. **Our own `_headers`** set `Cache-Control: no-cache` on `/player/_framework/*` and `/AppBundle/_framework/*`. That came from #1804, to fix SRI integrity failures on redeploy — none of those 190-odd filenames are content-hashed and AOT output isn't byte-reproducible, so a browser could pair a cached `dotnet.boot.js` (the manifest of per-file SHA-256 hashes) with a freshly-fetched binary from a later deploy and get a hash mismatch.
2. **Cloudflare Pages' default** for everything else is `public, max-age=0, must-revalidate` — including SvelteKit's already content-hashed `/_app/immutable/*`.
3. **The .NET loader does it too.** `dotnet.js` passes `cache: "no-cache"` to `fetch()` for every asset unless `disableNoCacheFetch` is set in `MonoConfig`, which bypasses freshness no matter what the CDN said. This one is easy to miss: fixing only the headers leaves the behaviour completely unchanged.

## The fix

Stamp a `?v=<version>` cache key onto every asset, then serve those paths `immutable`.

- `scripts/inject-version.mjs` stamps the 13 relative asset URLs in the markup (it already spliced the version in for the startup banner).
- `wasm-player.js` stamps what it fetches from JS — `playercore.htm`, `grid.js`, and `_framework/dotnet.js`. The runtime propagates that query to the JS modules and `dotnet.boot.js` by itself (`modulesUniqueQuery`), so `withResourceLoader` only needs to cover the assemblies, ICU data and `dotnet.native.wasm`; returning `null` for the rest avoids a doubled `?v=...?v=...`.
- `AppShell/src/lib/wasm.ts` does the same for the editor's AppBundle, off `PUBLIC_APPSHELL_VERSION`.
- `withConfig({ disableNoCacheFetch: true })` on both, without which the headers accomplish nothing.

This also fixes the integrity bug from #1804 properly rather than suppressing it. The version comes from `index.html`, which stays on `max-age=0, must-revalidate`, so manifest and binaries always carry the same version and can never come from different deploys — which is exactly what the blanket `no-cache` was buying, at the cost of ~190 revalidations per load.

## Measured

Warm reload of the Release AppBundle against a local server replicating the new header rules exactly:

| | network requests | from disk cache | 304s | transferred |
|---|---|---|---|---|
| before (live site) | 213 | 0 | 203 | 93 KB |
| after | **8** | 205 | 8 | ~0 |

Nothing under `_framework/` touches the network on a reload. The remaining 8 are `index.html`, the game file, and the six `lib/images/*` — the last referenced by relative URL from inside `jquery-ui.min.css`, so they can't be stamped and are deliberately left off the immutable list.

Isolating cause 3, with versioned URLs and immutable headers already in place but `disableNoCacheFetch: false`: **195 network requests, 195 × 304**. So the header change on its own would have been worth essentially nothing.

## On the safety of `immutable`

An `immutable` response to an *un*versioned URL is pinned in visitors' caches for a year and no later deploy can recall it, so the rules are scoped to paths where every request provably carries a version:

- `_framework/` is matched by extension (`*.wasm`, `*.dat`, `*.js`) rather than a blanket `/*`, so WasmEditor's `dotnet.native.js.symbols` — the one asset the runtime doesn't route through `loadBootResource`, so it can't be stamped — falls through to the default.
- `/player/index.html` and `/player/lib/images/*` are left out for the same reason.
- The `_headers` pattern semantics this relies on were confirmed against the live deployment rather than assumed: a `*` spans `/` (the existing `/*.js` rule demonstrably reaches `/player/lib/paper.js`), and a match is anchored at both ends (that same rule does *not* catch `dotnet.native.js.symbols`, which is served `application/octet-stream`).

`/_app/immutable/*` is a free win regardless — Vite already content-hashes those filenames.

## Test plan

- [x] `dotnet test --configuration Release` — all green
- [x] `npm run lint` and `npm run check` in `src/AppShell`
- [x] All 47 scripts flagged by `find-affected-tests.mjs` for the `wasm.ts` change — 47/47 pass
- [x] New `verify-wasmplayer-asset-versioning.mjs` guards the invariant (210 assets, all stamped exactly once, no SRI errors, `disableNoCacheFetch` on) — negative-tested by unstamping `grid.js`/`chrome.css` and by flipping `disableNoCacheFetch`, both correctly fail
- [x] `export-embedded.mjs` still produces a working single-file export — its two string replacements would otherwise have silently stopped matching once the tags gained a query, so they now tolerate it and throw on a miss
- [x] Dev server, Vite AppBundle middleware and the Electron loopback server all strip the query before resolving a path
- [ ] After merge: confirm `_framework` responses carry `immutable`, and that a redeploy doesn't produce integrity errors for a visitor holding the previous version

### DIFF
--- a/.github/workflows/deploy-play.yml
+++ b/.github/workflows/deploy-play.yml
@@ -69,11 +69,72 @@ jobs:
           cp -r src/WasmEditor/bin/Release/net10.0/browser-wasm/AppBundle/. deploy/AppBundle/
           cp -r src/WasmPlayer/bin/Release/net10.0/browser-wasm/AppBundle/. deploy/player/
           cat > deploy/_headers <<'EOF'
-          /player/_framework/*
-            Cache-Control: no-cache
+          # Cloudflare Pages' default for a static asset is
+          # `max-age=0, must-revalidate`, so without these rules a warm reload
+          # of /player spends ~210 conditional requests to move ~90 KB. Every
+          # path listed below is only ever requested with a `?v=<release>`
+          # cache key on it (WasmPlayer: inject-version.mjs stamps the markup,
+          # wasm-player.js the assets it fetches from JS; AppShell: wasm.ts) or
+          # is already content-hashed by Vite (/_app/immutable), so a deploy
+          # always produces a fresh URL and these can be cached indefinitely.
+          #
+          # For _framework/* the version is also what makes this *safe*, and it
+          # replaces the blanket `Cache-Control: no-cache` that used to sit
+          # here: those filenames aren't content-hashed and AOT output isn't
+          # byte-reproducible, so a browser pairing a cached dotnet.boot.js
+          # (the manifest of per-file SHA-256 hashes) with a binary from a
+          # later deploy got an SRI failure. Stamping both with the same
+          # version means they can't come from different deploys.
+          #
+          # Add nothing here that isn't requested with that version — an
+          # `immutable` response to an unversioned URL cannot be busted by any
+          # later deploy, by us or by the visitor. Three things are deliberately
+          # left on the default `must-revalidate` for exactly that reason:
+          # /player/index.html (which is what delivers each new version in the
+          # first place), /player/lib/images/* (referenced by relative URL from
+          # inside jquery-ui.min.css, so never stamped), and
+          # /AppBundle/_framework/dotnet.native.js.symbols (WasmEditor sets
+          # WasmEmitSymbolMap, and the runtime doesn't route that one asset
+          # through the resource loader that stamps all the others) — hence
+          # matching _framework/ by extension below rather than with a blanket
+          # `/*`, so the symbols file falls through to the default.
+          #
+          # Pattern semantics, confirmed against the live deployment rather
+          # than assumed: a `*` spans `/` (the `/*.js` rule below reaches
+          # /player/lib/paper.js), and a match is anchored at both ends (that
+          # same rule does *not* catch dotnet.native.js.symbols).
+          /player/_framework/*.wasm
+            Cache-Control: public, max-age=31536000, immutable
 
-          /AppBundle/_framework/*
-            Cache-Control: no-cache
+          /player/_framework/*.dat
+            Cache-Control: public, max-age=31536000, immutable
+
+          /AppBundle/_framework/*.wasm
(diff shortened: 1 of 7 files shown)


alexwarren:
Follow-up for #2186 is stacked on this branch: #2191. Merging this one will retarget it to `main`.

PR #2216, opened 2026-09-08
feat(AppShell): deep-link script commands into the documentation

alexwarren:
Follow-up to #2214, which added the baseline documentation links. This adds *contextual* help: each script command can open the reference page describing it.

## Affordances

- **ScriptEditor** — a `?` joining the existing hover cluster on a script row, plus a **Help** button in the single-selection toolbar as the touch path. That's the same split Move up / Move down already use, since the row's own buttons are deliberately hover-only at every nesting level.
- **AddScriptModal** — a **Learn more** link in the footer, tracking whichever command is highlighted.

Commands with no reference entry — syntax like `=` and `//`, and the `JS.` prefix — show no affordance at all rather than linking to a 404.

## Generated, not hand-maintained

The docs already use `## Name` headings whose anchors match the editor's `<appliesto>` keyword, so the mapping is mechanical. `site/scripts/build-docs-index.mjs` derives it and emits `src/AppShell/src/lib/docs-index.generated.ts`:

| Keyword class | Mapped |
|---|---|
| `(function)Foo` commands | 88 / 88 |
| Plain script keywords | 37 / 49 |

The 12 unmapped plain keywords are element types (`object`, `exit`, `timer`, …), which belong to the element reference rather than the function reference — deliberately out of scope here.

Three functions exist on both the gamebook and user-interface pages (`AddPageLink`, `HasSeenPage`, `RemovePageLink`). The index records both targets and the caller picks with `isGamebook`. The generator spells out both sides explicitly and throws if either heading disappears — relying on readdir order silently resolved all three to `functions/gamebook` (g sorts before u).

`check-docs-index` runs in CI next to the existing `check-function-docs`. It's a sibling in more than placement: that script already fails the build when a Core.aslx function has no `## ` heading, which is what keeps this index complete by construction.

## Plumbing

`EditorDefinition.AppliesTo` is promoted onto `IEditorDefinition` and carried through `ScriptNodeData`, so the client can identify an existing script row's command. The Add Script modal already had it via `ScriptCommandInfo.keyword`.

## ⚠️ questviva.com currently 404s on `/functions/*`

Worth knowing before merging: **88 of the 125 links target `/functions/*` pages that are not currently served in production.** `https://questviva.com/functions/objects/` returns the site's own 404 page, as does every other function page — even though they're listed in the live `sitemap-0.xml` and linked from the live site's own sidebar. `/scripts/` (the other 37) works fine.

This is pre-existing and unrelated to this PR: those pages were added in #2108, `npm run build` in `site/` produces all 18 of them locally, and the live nav already links to them. It looks like a partial or stale deploy of the docs site. Nothing here needs changing, but the deep links won't be useful until questviva.com is redeployed.

## Verification

- `dotnet build` + full `dotnet test` (399 tests across 5 projects) pass
- `npm run lint` / `npm run check` clean
- New `tests/e2e/verify-appshell-script-docs-links.mjs` covers both page shapes (`msg` → `/scripts/`, `MoveObject` → `/functions/objects/`), both affordances, and the negative case
- All 52 scripts in the union of `find-affected-tests.mjs` output plus `verify-appshell-i18n-pseudoloc.mjs` pass


alexwarren:
Rebased onto `main` (past #2218 and #2219) and the generated index regenerated for the new `/reference/functions/` path.

`build-docs-index.mjs` read the old location, so without this the 88 function deep-links would have pointed at a path that now only answers via a redirect. The section slug is a single `FUNCTIONS_SECTION` constant now rather than being spelled out in four places, so a future move is a one-line change.

**Verified against the live site**, now that both fixes are deployed — all 127 distinct link targets, every page and every anchor:

```
/reference/functions/attributes/            10 link(s)  0 bad
/reference/functions/core/                  11 link(s)  0 bad
/reference/functions/gamebook/               9 link(s)  0 bad
/reference/functions/general/                1 link(s)  0 bad
/reference/functions/internal-core/         14 link(s)  0 bad
/reference/functions/objects/               10 link(s)  0 bad
/reference/functions/string/                 2 link(s)  0 bad
/reference/functions/timers-turnscripts/    13 link(s)  0 bad
/reference/functions/user-interface/        22 link(s)  0 bad
/scripts/                                   35 link(s)  0 bad

TOTAL links checked: 127   broken anchors: 0
```

Every page returns 200 and every anchor exists on the page it points at. The ⚠️ caveat in the PR description above is resolved — that's what #2219 fixed.

Also re-ran: `check-docs-index`, `npm run lint` (site + AppShell), `npm run check` (svelte-check), and `verify-appshell-script-docs-links.mjs` against a local dev server — all 5 assertions pass, with the `MoveObject` expectation updated to the new path.

One conflict came up in the rebase (`site/package.json`, where #2218 and this PR both added npm scripts after `check-function-docs`); resolved by keeping both.

alexwarren:
Docs-content follow-ups from the review of this PR are tracked in #2220 — the six addable commands with no help link (`//`, `=`, `()`, `=>`, `JS.`, `HidePreviousTurnOutput`), and the thin reference entries where clicking through teaches you little. No change needed here; the plumbing picks both up automatically as the docs improve.

Issue #101, opened 2014-07-06
Beta 2 Editor: Limited container"It is full" message should be customizable


alexwarren:
It would be nice to expose this as a field we can customize, along the lines
of the "it contains" fields, etc.  
## Work Item Details

**Original CodePlex Issue:**    [Issue 725](http://quest.codeplex.com/workitem/725)
**Status:** Closed
**Reason Closed:**  Unassigned
**Assigned to:**    Unassigned
**Reported on:**    Jun 22, 2011 at 10:44 PM
**Reported by:**    xordevoreaux
**Updated on:** Jul 6, 2011 at 8:27 AM
**Updated by:** alexwarren
**Closed on:**  Jul 6, 2011 at 8:27 AM

Issue #100, opened 2014-07-06
"Set frame picture" in game start script doesn't work


alexwarren:
Add a "set frame picture" command to a game start script - no picture is
displayed.  
## Work Item Details

**Original CodePlex Issue:**    [Issue 724](http://quest.codeplex.com/workitem/724)
**Status:** Closed
**Reason Closed:**  Unassigned
**Assigned to:**    Unassigned
**Reported on:**    Jun 22, 2011 at 9:53 AM
**Reported by:**    alexwarren
**Updated on:** Jun 22, 2011 at 5:38 PM
**Updated by:** alexwarren
**Closed on:**  Jun 22, 2011 at 5:38 PM

PR #2217, opened 2026-09-08
fix(Site): stop Cloudflare Pages claiming the /functions/ docs section

alexwarren:
Every page under `https://questviva.com/functions/` currently returns the site's own 404 — all 18 function reference pages plus the section index — while every other section serves fine from the same build.

## It isn't a stale deploy

- `howto/world/exits.md` and `elements/index.md` were added in **the same commit** (`46fb33b2`) as `functions/objects.md`, and both serve 200
- the deployed `sitemap-0.xml` lists `https://questviva.com/functions/objects/` and its siblings
- the live sidebar links to `/functions/`
- the deployed Pagefind index reports `page_count: 167`, identical to a local build

Three separate artifacts of the deployed build reference these pages. They're built and uploaded; they just aren't served.

It also isn't `_redirects`: 685 static rules, one wildcard for a different hostname, nothing with `/functions` as a source, and the 404 carries no `Location` header.

## Cause

Cloudflare Pages reserves a top-level `functions` directory for [Pages Functions](https://developers.cloudflare.com/pages/functions/). The docs' function reference lives at `/functions/`, so the build emits `dist/functions/` — same name.

The name is the only variable. Nesting isn't involved (`/howto/world/exits/` is three levels deep and fine), and `/functions/objects/index.html` requested directly 404s exactly like a nonexistent path.

## The change

`public/_routes.json` declares what was already true — this site has no Functions, so no request should invoke one. Astro copies `public/` verbatim to the root of `dist/`; verified in a local build.

I used `"include": []` rather than the `"include": ["/*"], "exclude": ["/functions/*"]` form. Both express the intent, but if Pages *did* compile something out of that directory, pointing `/*` at it would put every request on a Worker this project never meant to have. Excluding everything can't misroute anything.

## How we'll know

This can only be validated by deploying — I can't see the Pages project settings.

- **Worked:** `/functions/` and `/functions/objects/` return 200 after the deploy.
- **Didn't:** they still 404, which would mean the files are dropped before the asset upload rather than shadowed by routing. The fallback is then to move the section to a non-colliding path (e.g. `/reference/functions/`) — more invasive, since ~400 rules in `_redirects` point at `/functions/...`.

Worth a look at the Pages build log either way: if it mentions compiling Functions, that confirms the mechanism outright.

## Related

[#2216](https://github.com/textadventures/quest/pull/2216) generates editor deep-links into the docs, and 88 of its 125 links target `/functions/*`. They stay dead until this is resolved. That PR needs no change if this works, and only a regenerated index if the section ends up moving instead.

### DIFF
--- a/src/AppShell/src/components/HomeHeader.svelte
+++ b/src/AppShell/src/components/HomeHeader.svelte
@@ -7,6 +7,8 @@
     import GithubIcon from "$components/GithubIcon.svelte";
     import DownloadButton from "$components/DownloadButton.svelte";
     import SettingsIcon from "@lucide/svelte/icons/settings";
+    import BookOpen from "@lucide/svelte/icons/book-open";
+    import { DOCS_URL } from "$lib/docs-links";
 
     let { forceDark = false }: { forceDark?: boolean } = $props();
 
@@ -23,6 +25,16 @@
                 <DownloadButton compact />
             </div>
         {/if}
+        <a
+            href={DOCS_URL}
+            target="_blank"
+            rel="noopener noreferrer"
+            class="home-header-link"
+            title={t("homeHeader.documentation")}
+            aria-label={t("homeHeader.documentation")}
+        >
+            <BookOpen size={18} />
+        </a>
         <a
             href="https://textadventures.co.uk/community/discord"
             target="_blank"
--- a/src/AppShell/src/components/Toolbar.svelte
+++ b/src/AppShell/src/components/Toolbar.svelte
@@ -43,11 +43,13 @@
     import Ellipsis from "@lucide/svelte/icons/ellipsis";
     import FileCode from "@lucide/svelte/icons/file-code";
     import SettingsIcon from "@lucide/svelte/icons/settings";
+    import BookOpen from "@lucide/svelte/icons/book-open";
     import DiscordIcon from "$components/DiscordIcon.svelte";
     import GithubIcon from "$components/GithubIcon.svelte";
     import DropdownMenu from "$components/DropdownMenu.svelte";
     import type { DropdownMenuItem } from "$components/DropdownMenu.svelte";
     import { isNarrow } from "$lib/layout.svelte";
+    import { DOCS_URL } from "$lib/docs-links";
 
     const DISCORD_URL = "https://textadventures.co.uk/community/discord";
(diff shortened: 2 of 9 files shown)


alexwarren:
Tested this on the PR's own Cloudflare Pages preview — **the `_routes.json` approach does not work**, so I've reset the branch and am closing this. No production deploy needed to find out; thanks to whoever set up preview deployments.

## `_routes.json` doesn't help

On the preview built from `60977c` (which had the file), `/functions/`, `/functions/objects/` and `/functions/user-interface/` all still returned 404, while `/scripts/`, `/intro/` and `/howto/world/exits/` returned 200. The preview's own `sitemap-0.xml` listed the functions pages and its Pagefind index reported the same `page_count: 167` as a local build — so they were built and uploaded, and simply not served.

Since previews are on `*.pages.dev`, this also rules out any zone-level rule on the questviva.com domain.

## What the cause actually is

I put two byte-identical files in the same build, differing only in directory name:

| Path | Result |
|---|---|
| `/functionsx/probe.txt` | 200, correct content |
| `/functions/probe.txt` | **404** |

Cloudflare Pages excludes a top-level `functions` directory from static assets **by name**, and `_routes.json` cannot override it — the exclusion happens at asset-manifest time, not routing. The original diagnosis was right about the mechanism and wrong about the lever.

A further probe showed `/reference/functions/probe.txt` → **200**, so the reservation applies only at the root of the build output. Nesting the section is a viable fix.

## Separate, larger problem found along the way

While testing whether redirect *sources* under `/functions/` still fire, I found that **only the first 100 rules in `_redirects` are honoured. The remaining 585 are silently ignored** — on production as well as previews.

The cutoff is exact: with a probe rule inserted at position 1, original rule #100 (`/switchable`) redirects and #101 (`/tabs_for_types`) 404s. Confirmed against questviva.com directly:

```
/speak_to        301   (rule #99)
/switchable      301   (rule #100)
/transcript      404   (rule #104)
/js/eval         404   (rule #150)
/attributes/look 404   (rule #300)
```

So most of the legacy Quest 5 URL redirects added in #2108 have never actually worked in production. That's independent of the `/functions/` issue and probably the more impactful of the two.

Closing this — the fix needs to be the rename instead, and the redirect limit wants handling separately.

Issue #19, opened 2014-07-06
Timers


alexwarren:
Add the ability to create timers.  

Also I think it would be useful to make "one-off" timers, a bit like
setTimeout in Javascript, so you can trigger some script after a certain
amount of time has elapsed.  
## Work Item Details

**Original CodePlex Issue:**    [Issue 360](http://quest.codeplex.com/workitem/360)
**Status:** Closed
**Reason Closed:**  Unassigned
**Assigned to:**    Unassigned
**Reported on:**    Oct 9, 2010 at 12:34 PM
**Reported by:**    alexwarren
**Updated on:** Jun 17, 2011 at 11:55 AM
**Updated by:** alexwarren
**Closed on:**  Jun 17, 2011 at 11:55 AM

PR #2219, opened 2026-09-08
fix(Site): move the function reference out of the reserved /functions/ path

alexwarren:
Stacked on #2218 (the redirect-capacity fix) — most of the `/functions/...` legacy rules sit past `_redirects`' 100-rule cutoff, so they'd be dead on arrival without it. **Review #2218 first**; this PR's diff is against that branch.

Every page under `questviva.com/functions/` currently returns a 404 — all 18 reference pages plus the section index — while the same build serves every other section fine.

## Cause, proven

Cloudflare Pages reserves a top-level `functions` directory for Pages Functions and excludes it from static assets **by name**. Two byte-identical files in one preview build:

| Path | Result |
|---|---|
| `/functionsx/probe.txt` | 200, correct content |
| `/functions/probe.txt` | **404** |

`_routes.json` can't override it (tried in #2217, now closed) — the exclusion happens at asset-manifest time, not routing. A further probe showed `/reference/functions/probe.txt` → 200, so the reservation applies only at the **root** of the build output.

## The move

`/functions/` → `/reference/functions/`, matching the sidebar section these pages already live in ("Language Reference").

The other Language Reference pages (`/scripts/`, `/attributes/`, `/elements/`, `/types/`) stay top-level. That's a deliberate asymmetry: moving them too would be a wider IA change with no functional reason behind it. If you'd rather have the whole section under `/reference/`, that's a clean follow-up now that redirects work.

## Why the old namespace redirects from `_redirects`

An Astro redirect stub built at `/functions/...` would be excluded exactly like everything else there. But redirect *rules* still fire for those paths (verified with a probe rule on the preview), so two dynamic rules forward the whole namespace:

```
/functions        /reference/functions        301
/functions/*      /reference/functions/:splat 301
```

The 317 fine-grained legacy mappings then take over inside the new namespace — e.g. `/functions/corelibrary/addtoinventory` → `/reference/functions/corelibrary/addtoinventory` → `/reference/functions/core#addtoinventory`.

## Also

Drops `src/content/docs/reference/example.md` — unused Starlight scaffold ("A reference page in my new Starlight docs site"), never in the sidebar, which would have sat oddly beside real reference content. Redirected to `/intro`.

## Verification

- `check-function-docs` and `check-redirects` both pass (783 redirects resolve, no chains, no broken anchors)
- 36 markdown files had internal links rewritten; **zero** `href="/functions…"` remain anywhere in the built HTML
- `dist/functions/` no longer exists, so nothing collides with the reserved name
- 335 pages under `reference/functions/` (18 real + 317 legacy stubs); legacy stubs verified to forward correctly
- `npm run lint` clean

Preview-deployment results to follow on this PR.

## Follow-up

#2216's generated docs index points at `/functions/...`; it just needs regenerating (`node site/scripts/build-docs-index.mjs`) once this lands.


alexwarren:
## Preview-deployment results — both fixes confirmed working

Verified against `https://feat-site-functions-rename.quest-1qv.pages.dev` (this PR's Cloudflare preview, which includes #2218).

**The rename resolves the 404s.** All 18 reference pages return 200:

```
/reference/functions/ + all 17 category pages   ->  200
```

**The old namespace forwards correctly** via the two dynamic `_redirects` rules:

```
/functions                  301 -> /reference/functions
/functions/objects          301 -> /reference/functions/objects
/functions/user-interface/  301 -> /reference/functions/user-interface/
```

**Legacy deep links resolve end-to-end**, following the `_redirects` splat and then the Astro stub (chains resolved manually, since curl doesn't follow meta refresh):

```
/functions/corelibrary/addtoinventory  ->  /reference/functions/core#addtoinventory       200
/functions/allobjects                  ->  /reference/functions/scope#allobjects          200
/functions/listworn                    ->  /reference/functions/clothing#listwornfor      200
/functions/fn-scope                    ->  /reference/functions/scope                     200
```

**And #2218's fix is confirmed on the same preview** — these all 404 in production today:

```
/transcript         ->  /howto/world/transcript    200
/js/eval            ->  /js#eval                   200
/attributes/look    ->  /attributes#look           200
/cloak_of_darkness  ->  /cloak-of-darkness         200
/reference/example  ->  /intro                     200
```

`/cloak_of_darkness` is worth calling out: it's rule #19, so it's live in production right now and 301s to a page that 404s. It lands correctly here.

Issue #2227, opened 2026-09-09
Sign on to the Treaty of Babel


dfabulich:
https://babel.ifarchive.org/babel.html

Quest should be a format on this list.

1. Add `UUID://...//` zip comment when publishing `.quest` files
2. The editor should embed an `iFiction` document in `.quest` archives
    * iFiction expects releases/attached/release tag to be a non-negative integer; Quest will presumably need to have a `<versioncode>` element in addition to the `<version>` string
3. We should teach the Babel tool how to extract an IFID for legacy `.quest` archives and `.aslx` files (extract `.quest` as a ZIP, find the `asl/game/gameid` element in the ASLX, using the ASLX MD5 for ASLX files without a `<gameid>` element)
    * … and how to synthesize an iFiction document from ASLX
    * … and how to synthesize an iFiction document from ASL
    * … and how to decompile CAS files to ASL
4. We should file PRs on https://github.com/iftechfoundation/ifarchive-if-specs/blob/main/Babel-Treaty.md and https://github.com/iftechfoundation/babel-tool to add support for Quest games
5. Bibliographic ASLX elements should be documented in https://docs.textadventures.co.uk/quest/elements/game.html
     * `<gameid>`
    * `<version>`
    * `<versioncode>` (presuming we add that)
    * `<firstpublished>`
    * `<subtitle>`
    * `<author>`
    * `<description>`
    * `<category>`
    * `<cover>`
    * `<difficulty>` (legacy, but should still be documented)
    * `<cruelty>` (legacy, but should still be documented)

dfabulich:
OK, I've filed four PRs on this repository to prepare for integration with the Treaty of Babel.

* https://github.com/textadventures/quest/pull/2229
* https://github.com/textadventures/quest/pull/2230 
* https://github.com/textadventures/quest/pull/2231
* https://github.com/textadventures/quest/pull/2237

I've also filed a documentation PR:
* https://github.com/textadventures/quest/pull/2232

And I've filed a draft PR to update the Treaty:
* https://github.com/iftechfoundation/ifarchive-if-specs/pull/29

(I think we should merge the Quest PRs above before merging the PR to update the treaty)

And here's a draft PR to teach the Babel tool how to access all the metadata:
* https://github.com/iftechfoundation/babel-tool/pull/45

alexwarren:
[...]

PR #2229, opened 2026-09-09
feat: add numeric `<versioncode>` field for games

dfabulich:
Introducs a Babel-friendly integer version code alongside the freeform Version string, with Setup-tab editing, VERSION command output, and templates/i18n support.

* Apropos #2227

image

image

### DIFF
--- a/tests/e2e/verify-appshell-versioncode.mjs
+++ b/tests/e2e/verify-appshell-versioncode.mjs
@@ -1,0 +1,83 @@
+// Verifies the Setup-tab Version Code field (<versioncode>): number input on the
+// same horizontal row as Version, default 1 from defaultgame, and
+// persisted as an int in the game XML.
+//
+// Run against a dev server started with:
+//   PUBLIC_SHOW_HOME=true npm --prefix src/AppShell run dev -- --port 5180
+import { chromium } from './lib/tracked-chromium.mjs';
+
+const baseUrl = process.argv[2] || 'http://localhost:5180';
+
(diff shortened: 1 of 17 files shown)


src/Engine/Core/Templates/English.template
```diff
@@ -6,6 +6,7 @@
   <game name="$NAME$">
     <gameid>$ID$</gameid>
     <version>1.0</version>
+    <versioncode type="int">1</versioncode>
```
alexwarren: Games being edited that were created prior to this template update won't have this attribute at all - it might be a good idea to add this to the `defaultgame` type in `CoreTypes.aslx` too to set a default value. In fact, then we wouldn't need the template updates as this would just be set in that one place.

tests/e2e/verify-appshell-versioncode.mjs
```diff
@@ -0,0 +1,83 @@
+// Verifies the Setup-tab Version Code field (<versioncode>): number input on the
```
alexwarren, line 1: This file also needs adding to `.github/workflows/e2e.yml` so it doesn't get missed when the e2e tests run.

alexwarren (changes requested):
Looks good! Just a couple of tweaks needed.

tests/e2e/verify-appshell-versioncode.mjs
alexwarren: Ah I see you just did that

alexwarren (approved):

PR #2230, opened 2026-09-09
feat: embed `metadata.iFiction` in published `.quest` packages

dfabulich:
* Apropos #2227 
* Apropos https://github.com/iftechfoundation/ifarchive-if-specs/pull/29
* Apropos https://github.com/iftechfoundation/babel-tool/pull/45

Here's an example output file.

* It's using a new `<format>quest</format>`
* And we've defined a new `<quest>` format-specific tag, per [Treaty of Babel section 5.10](https://babel.ifarchive.org/babel.html#the-format-specific-tags). Within the `<quest>` element:
    * `<style>` can be "Text Adventure" or "Gamebook"
    * `<coverleafname>` is the name of the cover art file in the `.quest` ZIP archive. (Strangely, Babel has no standardized way of identifying the actual file name of the cover art!)
    * `<version>` is the `<version>` version string from the ASLX.
* The `releases/attached/release/version` is the ASLX `<versioncode>`, introduced in PR 2229.

```
<?xml version="1.0" encoding="UTF-8"?>
<ifindex version="1.0" xmlns="http://babel.ifarchive.org/protocol/iFiction/">
  <story>
    <identification>
      <ifid>1652E7EC-2D7D-4B59-95ED-E65D79EDF257</ifid>
      <format>quest</format>
    </identification>
    <bibliographic>
      <title>Test</title>
      <author>Dan Fabulich</author>
      <headline>subtitle</headline>
      <genre>Educational</genre>
      <firstpublished>2026</firstpublished>
      <description>This is a test. &lt;3<br/>multiple paragraphs<br/>line 1<br/>line 2 with no paragraph break</description>
      <language>en</language>
    </bibliographic>
    <cover>
      <format>png</format>
      <height>512</height>
      <width>512</width>
    </cover>
    <quest>
      <style>Text Adventure</style>
      <version>1.0</version>
      <coverleafname>black-cover.png</coverleafname>
    </quest>
    <releases>
      <attached>
        <release>
          <version>1</version>
          <releasedate>2026-09-09</releasedate>
          <compiler>Quest Viva</compiler>
          <compilerversion>6.0.0-beta.59</compilerversion>
        </release>
      </attached>
    </releases>
    <colophon>
      <generator>Quest Viva</generator>
      <generatorversion>6.0.0-beta.59</generatorversion>
      <originated>2026-09-09</originated>
    </colophon>
  </story>
</ifindex>
```

### DIFF
--- a/src/Engine/GameLoader/IFictionWriter.cs
+++ b/src/Engine/GameLoader/IFictionWriter.cs
@@ -1,0 +1,425 @@
+using System.Globalization;
+using System.Net;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Xml;
+using QuestViva.Common;
+
+namespace QuestViva.Engine.GameLoader;
+
+/// <summary>
+///     Builds a Treaty of Babel iFiction record from a WorldModel's bibliographic fields,
+///     for embedding as <c>metadata.iFiction</c> inside a published <c>.quest</c> package.
+/// </summary>
+internal static class IFictionWriter
+{
+    private static readonly Regex Whitespace = new(@"\s+", RegexOptions.Compiled);
+    private static readonly Regex HtmlTag = new(@"<[^>]+>", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+    private static readonly Regex BrTag = new(@"<br\s*/?>", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+    private static readonly Regex ClosePTag = new(@"</p\s*>", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+    private static readonly Regex OpenPTag = new(@"<p\b[^>]*>", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+    private static readonly Regex FirstPublishedYear = new(@"^\d{4}$", RegexOptions.Compiled);
+    private static readonly Regex FirstPublishedDate = new(@"^\d{4}-\d{2}-\d{2}$", RegexOptions.Compiled);
+
+    public static string Write(WorldModel worldModel,
+        IEnumerable<WorldModel.PackageIncludeFile>? includeFiles)
+    {
+        var game = worldModel.Game.Fields;
+        var gameId = game.GetString("gameid");
+        if (string.IsNullOrWhiteSpace(gameId))
+        {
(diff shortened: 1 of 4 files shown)


alexwarren:
Annoying how stacked PRs don't play nicely with "Squash and Merge" - this will need rebasing

src/Engine/GameLoader/IFictionWriter.cs
```diff
@@ -0,0 +1,450 @@
...
+            return null;
+        }
+
+        var format = CoverFormatFromFilename(coverName);
+        if (format == null)
+        {
+            return null;
+        }
+
+        var position = file.Content.CanSeek ? file.Content.Position : 0;
```
alexwarren: If `file.Content` isn't seekable, reading the header here consumes the start of the cover image and there's no way to rewind it — `WriteZip` then copies a truncated image into the package. Every current caller passes a `MemoryStream` so it can't happen today, but `EditorController.Publish` is public and accepts any `Stream`.

Suggest bailing out early with `if (!file.Content.CanSeek) return null;` (no cover metadata rather than a corrupted cover), which also makes the non-seekable skip loop and `skipBuffer` in `TryReadJpegSize` dead code that can be deleted.

src/Engine/GameLoader/IFictionWriter.cs
```diff
@@ -0,0 +1,450 @@
...
+            _ => false
+        };
+    }
+
+    private static bool TryReadPngSize(Stream stream, out int width, out int height)
+    {
+        width = 0;
+        height = 0;
+        Span<byte> header = stackalloc byte[24];
+        if (stream.Read(header) < 24)
```
alexwarren: `Stream.Read` is allowed to return fewer bytes than requested even when more are available, so a short read here would wrongly fail. `stream.ReadAtLeast(header, 24, throwOnEndOfStream: false) < 24` is the safer form. (No practical difference with `MemoryStream`.)

alexwarren:
Looks good - I was curious if .NET had anything to do the image size detection itself, so I asked Claude. TL;DR - no, so having our own code here works best. But it did spot a couple of small things so I got it to just add comments directly itself.

alexwarren (approved):

PR #2286, opened 2026-09-12
docs: teach the synchronous input forms instead of the callback commands

alexwarren:
Several docs pages still taught `get input { }`, `ask { }`, `show menu { }` and `wait { }` — the four callback script commands that are no longer offered in the editor's script adder, now that `GetInput()`/`Ask()`/`ShowMenu()`/`request (Wait, "")` are re-enabled for v600 games.

This follows on from the What's new bullet in #2284, with the same framing: the callback forms still run, they're just not the way to write anything new.

### Pages rewritten

- **`howto/tasks/asking-a-question.md`** — the highest-priority one, since it's entirely about this. Rewritten around `GetInput()`:
  - "Multiple questions" no longer needs nesting, so the section becomes "here's a second question" plus a short explanation of why the old command needed nesting;
  - the riddle example's tail logic (putting the prompt and panes back) moves out of the block — a callback example can't just be search-and-replaced, since anything after the block used to run *before* the player had answered;
  - "Dozens of questions" drops the recursive `AskAQuestion` function for a `foreach` loop.
- **`howto/tasks/keeping-a-journal.md`** — the `USE JOURNAL` script.
- **`howto/scripting/blocks-and-scripts.md`** — the `ask { }` example is kept as the illustration of the non-blocking trap (that's what the section is for), but is now followed by the `Ask()` rewrite. A note at the top says which callback forms remain: `on ready`, the timeout functions, and the `Ask (...) { }` / `ShowMenu (...) { }` *function* forms, which are deliberately still addable for their inline-link rendering.
- **`scripts/index.md`** — all four entries stay (old games contain these commands and authors need to look them up); each gains a short "Superseded" note linking on to the synchronous form. No headings renamed, and `check-docs-index` passes. The `request` table's `Wait` row pointed at the `wait` command, which is now backwards.

### Also touched

`reference/functions/user-interface.md` — these are the pages the new links land on, and they said the opposite: `GetInput` and `WaitForKeyPress` carried stale "deprecated as of Quest 5.x / unsupported" banners, and `Ask`/`ShowMenu` documented only their callback form. Both now lead with the expression form and keep the callback form in a subsection. Not in the original scope, but leaving them would have meant shipping links that contradict themselves.

### Verification

Engine behaviour behind each rewritten example was checked against a `GameDriver` over a v600 game rather than assumed — `GetInput()` inside a `foreach`, two `GetInput()` calls in a row, chained `ShowMenu()` calls, `ShowMenu()` returning `""` (not null) on cancel, `Ask()` in an `if` with the tail running after, and `WaitForKeyPress` twice in one script.

`Question1-4.png` regenerated with `capture-asking-a-question.mjs`, updated for the new scripts. `Question2.png` happens to show both renderings of `GetInput()` — "player's typed input" where it's the whole value, and a plain expression field where it's wrapped in `CapFirst`.

From `site/`: `npm run build`, `check-editor-help-urls`, `check-docs-index`, `check-redirects`, `check-function-docs` and `lint` all pass.

### Not done

Three more pages use the still-addable `ShowMenu (...) { }` / `wait { }` callback forms and would benefit from the same pass, but they weren't in scope here and the menu story is more nuanced (the callback form is deliberately kept):

- `howto/rpg/character-creation.md` — already half-migrated, and now says `show menu` "unlike `GetInput()`, has to wait for the player to click an option, so the code that handles the answer has to go inside its callback block", which the `ShowMenu()` expression form makes untrue.
- `howto/tasks/showing-a-menu.md` — frames `show menu` as the popup and `ShowMenu` as the inline one; both forms of `ShowMenu` now exist and only the callback one is inline.
- `howto/scripting/using-dictionaries.md` — one `ShowMenu (...) { }` example.


alexwarren:
Pushed f92e8740 with the review changes.

- **`CapFirst` dropped** from the first two examples — you were right that it broke the template detection. `Question1.png`/`Question2.png` regenerated and both now show "player's typed input" selected. The "makes sure there is a capital at the start" sentence went with it. Capture script comment updated to record *why* the examples assign `GetInput()` directly, so nobody reintroduces a wrapper later.
- **Quest 5 note softened** to "may use the `get input` script command … that was the way to do this from Quest 5.4 onwards", since `GetInput()` was available before then.
- **Historical asides cut.** Gone from asking-a-question.md: the "this is the thing the old command got wrong" paragraph plus its nested `get input` example, the "with `get input { }` these lines would have had to go inside the block" note, and the "used to take a recursive function" note. The page is now just: here's how you ask a question, with one short note pointing at `get input` for anyone reading an existing game. blocks-and-scripts.md got the same trim, though the `ask { }` → `Ask()` comparison stays there since callback mechanics are that page's actual subject.

**On the popups — you're right to be worried, and it's worse than I'd assumed.** Verified in WasmPlayer against a v600 fixture:

| Form | Renders as |
|---|---|
| `Ask (q)` expression | jQuery UI modal, "Question" title bar, Yes/No buttons |
| `Ask (q) { }` callback | inline `1: Yes` / `2: No` links in the transcript |
| `ShowMenu (...)` expression | jQuery UI modal, `<select size=4>` listbox **plus a separate "Select" button** |
| `ShowMenu (...) { }` callback | inline numbered links in the transcript |
| `GetInput()`, `WaitForKeyPress` | normal prompt / Continue link — fine, nothing to fix |

So it's just the two that route through `IPlayer.ShowMenu`/`ShowQuestion` → `showMenu()`/`showQuestion()` in `player.js`. The menu one is the worst of it: picking one of four options takes two interactions, and on a 390px viewport the dialog sits at ~298px floating over the transcript rather than laying out for the screen. It also won't follow the player's dark theme.

Filed as #2287, with the repro fixture, the call chain, and a note that whichever way it goes (inline links vs. a properly themed modal) is a decision worth making before building. The docs pages that currently describe these as popups are listed there too, so they get revisited when it changes.

Worth saying: this is pre-existing — the `ask`/`show menu` script commands always went through the same dialogs. What's changed is that these are now the *recommended* path, so the ugliness is newly in front of people.

